PR #2, opened 2015-02-23
S3 delegation api

jonasfj:
See bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1127126


bin/server.js
```diff
@@ -73,7 +74,8 @@ var launch = function(profile) {
       context: {
         validator:      validator,
         Client:         Client,
-        azureAccounts:  JSON.parse(cfg.get('auth:azureAccounts'))
+        azureAccounts:  JSON.parse(cfg.get('auth:azureAccounts')),
+        sts:            new AWS.STS(cfg.get('aws'))
```
jhford: Is this only for a single region?  Does it matter that we're going to have instances in other regions as well?

routes/api/v1.js
```diff
@@ -424,6 +427,120 @@ api.declare({
...
+    "subject to change. Hence, you should always read the `expires` property",
+    "from the response, if you intent to maintain active credentials in your",
+    "application.",
+    "",
+    "Please notice that your `prefix` shouldn't start with slash `/`, it is",
+    "allowed for completeness, but it will result in two slashes both in the",
+    "scope and in the bucket. Therefore this pattern is strongly discouraged.",
+    "Also note that if your `prefix` doesn't end in a slash `/` you will not",
+    "be required to insert one. This is mainly a concern when assigning scopes",
+    "to users and doing this right will prevent poor behavior."
```
jhford: is there any case where we should allow `//` in the name?  Will S3 know how to serve such a file properly?

routes/api/v1.js
```diff
@@ -424,6 +427,120 @@ api.declare({
...
+
+  return this.sts.getFederationToken({
+    Name:               'TemporaryS3ReadWriteCredentials',
+    Policy:             JSON.stringify({
+      Version:          '2012-10-17',
+      Statement:[
+        {
+          Sid:            'ReadWriteObjectsUnderPrefix',
+          Effect:         'Allow',
+          Action:         (level === 'read-write' ? [
```
jhford: Multiline-ternary is really ugly.  Can we create an `action` list outside of this object definition and use it here instead?

routes/api/v1.js
```diff
@@ -424,6 +427,120 @@ api.declare({
...
+          Action: [
+            's3:GetBucketLocation'
+          ],
+          Resource: [
+            'arn:aws:s3:::{{bucket}}'
+              .replace('{{bucket}}', bucket)
+          ]
+        }
+      ]
+    }),
```
jhford: I'm not sure if all this stuff is correct... how sure are you with it?

routes/api/v1.js
jonasfj: yes.. let's assert and remove the assert if it's ever a problm.

bin/server.js
jonasfj: AWS STS credentials works globally: http://docs.aws.amazon.com/STS/latest/UsingSTS/Welcome.html

So that should be fine. Note at this point this is only for S3 access. Special case access like EC2 should be done with custom IAM users and custom policies.

routes/api/v1.js
jonasfj: Reasonably sure... :)

I'm not scared of all the `get` type actions. 

The only thing that is slightly _interesting_ is the prefix thing...
But all of this comes from: http://blogs.aws.amazon.com/security/post/Tx1P2T3LFXXCNB5/Writing-IAM-policies-Grant-access-to-user-specific-folders-in-an-Amazon-S3-bucke

Note, the tests check for both access and access denied. So it works :)

routes/api/v1.js
```diff
@@ -424,6 +427,120 @@ api.declare({
...
+    "be required to insert one. This is mainly a concern when assigning scopes",
+    "to users and doing this right will prevent poor behavior."
+  ].join('\n')
+}, function(req, res) {
+  var level   = req.params.level;
+  var bucket  = req.params.bucket;
+  var prefix  = req.params.prefix;
+
+  // Validate that a proper value was given for level
+  if (level !== 'read-write' && level !== 'read-only') {
```
jonasfj: @jhford,
So honestly this is the part that I'm most in doubt about.

Is it really smart to include `read-only` and `read-write` as parameters in the URL. Is that a nice design?

PR #5, opened 2015-07-06
Miscellaneous docstring edits

djmitche:
There are a few semantic changes here, based on the code; please verify?

### DIFF
--- a/routes/api/v1.js
+++ b/routes/api/v1.js
@@ -34,10 +34,10 @@ api.declare({
   title:      "Get Client Authorized Scopes",
   description: [
     "Returns the scopes the client is authorized to access and the date-time",
-    "where the clients authorization is set to expire.",
+    "when the clients authorization is set to expire.",
     "",
     "This API end-point allows you inspect clients without getting access to",
-    "credentials, as provide by the `getCredentials` request below."
+    "credentials, as provided by the `getCredentials` request below."
   ].join('\n')
 }, function(req, res) {
   return this.Client.load(req.params.clientId).then(function(client) {
@@ -68,7 +68,7 @@ api.declare({
   scopes:     ['auth:credentials'],
   title:      "Get Client Credentials",
   description: [
-    "Returns the clients `accessToken` as needed for verifying signatures.",
+    "Returns the client's `accessToken` as needed for verifying signatures.",
     "This API end-point also returns the list of scopes the client is",
     "authorized for and the date-time where the client authorization expires",
     "",
@@ -106,9 +106,9 @@ api.declare({
   scopes:     ['auth:credentials'],
   title:      "Get Client Information",
   description: [
-    "Returns all information about a given client. This end-point is mostly",
+    "Returns all information about a given client. This end-point is mostly for",
     "building tools to administrate clients. Do not use if you only want to",
-    "authenticate a request, see `getCredentials` for this purpose."
+    "authenticate a request; see `getCredentials` for this purpose."
   ].join('\n')
 }, function(req, res) {
   return this.Client.load(req.params.clientId).then(function(client) {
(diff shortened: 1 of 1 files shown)


routes/api/v1.js
```diff
@@ -142,13 +142,12 @@ api.declare({
...
-    "Create client with given `clientId`, `name`, `expires`, `scopes` and",
+    "Create a client with given `clientId`, `name`, `expires`, `scopes` and",
     "`description`. The `accessToken` will always be generated server-side,",
     "and will be returned from this request.",
     "",
-    "**Required scopes**, in addition the scopes listed",
-    "above, the caller must also posses the all the scopes that is given to",
-    "the client that is created."
+    "**Required scopes**: in addition the scopes listed",
+    "above, the `scopes` parameter must satisfy the caller's scopes."
```
jonasfj: I think "the `scopes` parameter must satisfy the caller's scopes" got flipped around.

```
  "**Required scopes**: in addition the scopes listed",
  "above, the caller must satisfy all scopes granted to the client."
```
or if you must:
```
  "**Required scopes**: in addition the scopes listed",
  "above, the caller must satisfy all scopes from the `scopes` property of the payload."
```

routes/api/v1.js
```diff
@@ -223,9 +222,9 @@ api.declare({
     "Modify client `name`, `expires`, `scopes` and",
     "`description`.",
     "",
-    "**Required scopes**, in addition the scopes listed",
-    "above, the caller must also posses the all the scopes that is given to",
-    "the client that is updated."
+    "**Required scopes**: in addition the scopes listed",
+    "above, the `scopes` parameter must satisfy the caller's scopes.  The",
+    "client's existing scopes need not satisfy the caller's scopes."
```
jonasfj: Again this is reversed...
It's the caller that must satisfy all the scopes assigned to the modified client. Also I'm not happy calling it the `scopes` parameter, it's the `scopes` property or `client.scopes` property, or `scopes` property of the payload. That's mostly nit...

However, direction of satisfiability is important.

jonasfj: note: `client.scopes <= caller.scopes` where `client` is the payload. Hence, caller must satisfy the scopes granted to the client.

routes/api/v1.js
```diff
@@ -439,20 +440,20 @@ api.declare({
...
     "application.",
     "",
-    "Please notice that your `prefix` may not start with slash `/`, it is",
-    "allowed on S3, but we forbid it here to discourage bad behavior.",
-    "Also note that if your `prefix` doesn't end in a slash `/` the STS",
-    "credentials will not require one to be to inserted. This is mainly a",
+    "Please note that your `prefix` may not start with slash `/`. Such a prefix",
+    "is allowed on S3, but we forbid it here to discourage bad behavior.",
+    "Also note that if your `prefix` doesn't end in a slash `/`, the STS",
+    "credentials will not require a trailing slash. This is mainly a",
```
jonasfj: Trailing slash is a different thing.
You prefix can be `my-folder/` in which case you can modify `my-folder/my-file.txt`, etc...

If the prefix doesn't have a trailing prefix, one will not be required to be inserted, and hence, if you have prefix `my-folder` you may modify `my-folder/my-file.txt` as before, but you may also modify `my-folder-NOT.txt` :)

So if you security relies on folder structures and not prefixes one must include trailing slash in the prefix.

routes/api/v1.js
djmitche: Yeah, I posted this before we talked in irc.  I'll revise.

routes/api/v1.js
djmitche: I think we're saying the same thing - "one will not be required to be inserted" is a double-passive form of "there is no need to insert one", where "one" refers to a trailing slash.  But I'll try to say it a different way.

djmitche:
OK, updated -- have another look?  Sorry again for my confusion!

routes/api/v1.js
```diff
@@ -223,9 +222,9 @@ api.declare({
     "Modify client `name`, `expires`, `scopes` and",
     "`description`.",
     "",
-    "**Required scopes**, in addition the scopes listed",
-    "above, the caller must also posses the all the scopes that is given to",
-    "the client that is updated."
+    "**Required scopes**: in addition the scopes listed",
+    "above, the `scopes` property must be satisfied by the caller's",
+    "scopes.  The client's existing scopes are not considered."
```
jonasfj: This is nit, but if we do have colon, I believe one should follow with a capital "In", not sure though... Either way, I was happy with comma and remains happy with colon :)

Note, `satisfied by the caller` is likely sufficient to imply satisfied by the caller's scopes. But again, it's nit picking...

PR #8, opened 2015-08-24
Remote signature validation

jonasfj:
Adds the API end-point needed for https://github.com/taskcluster/taskcluster-base/pull/26

And utilities used for load testing this...

See [load-test results here](https://docs.google.com/spreadsheets/d/1aU5qPO2xZw6xoAPEanjGpcx21gbcbOPSl2e_S8GXFoE/edit?usp=sharing)

### DIFF
--- a/auth/data.js
+++ b/auth/data.js
@@ -1,6 +1,7 @@
 var debug   = require('debug')('auth:data:client');
 var base    = require('taskcluster-base');
 var assert  = require('assert');
+var _       = require('lodash');
 
 /** Configure a client Entity subclass */
 var Client = base.LegacyEntity.configure({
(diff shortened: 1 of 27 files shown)


schemas/authenticate-hawk-response.yml
```diff
@@ -0,0 +1,45 @@
+$schema:  http://json-schema.org/draft-04/schema#
+title:                    "Hawk Signature Authentication Response"
+description: |
+  Response from a request to authenticate a hawk request.
+anyOf:
```
jonasfj: I'm not sure this schema is elegant... Perhaps I need to add a `status: "authorized" || "unauthorized"` to make it easy to distinguish between the two objects for `anyOf`. Thoughts?

Mostly I focus on keeping it future proof where we can support a different authentication scheme.

djmitche:
@gdestuynder you might want to have a look, too

schemas/authenticate-hawk-response.yml
djmitche: I think that would certainly make parsing the response easier -- `"message" in response` doesn't scan as "if the authentication failed"

schemas/authenticate-hawk-response.yml
```diff
@@ -0,0 +1,45 @@
...
+          read this property unless `hash` is provided and you want to validate
+          the payload hash. Additional values may be added in the future.
+        type:                 string
+        enum:
+          - hawk
+      hash:
+        description: |
+          Payload as extracted from `Authentication` header. This property is
+          only present if a hash is available. You are not required to validate
+          this hash, but if you do, please check `schema` to ensure that it's
```
djmitche: `s/schema/scheme/`

auth/data.js
```diff
@@ -78,6 +79,70 @@ Client.createClientLoader = function() {
...
+    return {
+      clientId:     client.clientId,
+      accessToken:  client.accessToken,
+      scopes:       client.scopes
+    };
+  };
+};
+
+
+/** Create caching client loader */
```
mrrrgn: This awesome, but I can't help but feel like a more descriptive name would be helpful to someone reading the code in the future. createCachedClientLoader ? #nit ;p

bin/load-test.js
```diff
@@ -0,0 +1,200 @@
...
+var hawk          = require('hawk');
+var taskcluster   = require('taskcluster-client');
+var assert        = require('assert');
+
+/** Launch server */
+var launch = async function(profile) {
+  debug("Launching with profile: %s", profile);
+
+  // Load configuration
+  var cfg = base.config({
```
mrrrgn: This load test is neat. Just to make things more DRY, any chance we wrap this base.config({...}) in a utility function? The chance of an env var's name changing may be small; but it would be nice to have one source of truth in any case.

schemas/authenticate-hawk-response.yml
mrrrgn: A status would be less ambiguous at a glance. :+1:

bin/load-test.js
jonasfj: Yes... config loading could be MUCH smarter... I hate how stupid it is... :)
It's really a foot gun and hard to get new people started, I moved my hack from f? to r?, and since you showed interest I figure you might like to review it:
https://bugzilla.mozilla.org/show_bug.cgi?id=1084661#c13

I haven't really seen much interest from others, and I can't really tell if the idea here (abusing YAML custom syntax) is the right or wrong amount of crazy :)

schemas/authenticate-hawk-response.yml
jonasfj: it does have status now...
`status: "auth-success"` or `status: "auth-failed"`

I just use enums to enforce a specific value, depending on the object form (ie. the anyOf branch)...

jonasfj: and yes, it's much clearer to do this, thanks! :)

auth/data.js
jonasfj: Will rename to `createCachedClientLoader` and the old one to `createLegacyClientLoader`... No more 2, 3 :)

jonasfj: I hate having bad names... :)

PR #9, opened 2015-09-04
Add expired test

jonasfj:
As you pointed out we needed a test like this... Thanks! :)

### DIFF
--- a/test/remotevalidation_test.js
+++ b/test/remotevalidation_test.js
@@ -216,7 +216,7 @@ suite("Remote Signature Validation", () => {
     var clientId = slugid.v4();
     var result = await helper.auth.createClient(clientId, {
       scopes:       ['myapi:*'],
-      expires:      new Date(1998, 1, 1), // far out in the future
+      expires:      new Date(1998, 1, 1), // far back in the past
       name:         "test-client",
       description:  "Client used by automatic tests, file a bug and delete if" +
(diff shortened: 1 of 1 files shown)


test/remotevalidation_test.js
```diff
@@ -191,4 +191,48 @@ suite("Remote Signature Validation", () => {
...
+      }
+    });
+    await myClient.resource();
+  });
+
+  test("auth with non-root user (expired)", async () => {
+    var clientId = slugid.v4();
+    var result = await helper.auth.createClient(clientId, {
+      scopes:       ['myapi:*'],
+      expires:      new Date(1998, 1, 1), // far out in the future
```
petemoore: I think this should say `// far back in the past`

jhford: yep!

jhford:
r+ from me, i'll let you deploy when it makes sense to do so!

djmitche:
@jonasfj does this still need to be landed?

djmitche:
Hah, well, these tests now fail -- we appear not to ever check whether a client is expired.

djmitche:
merged as part of #32

djmitche:
(the merge commit got flattened in a rebase, oops)

jonasfj:
awesome thanks for landing this....

PR #10, opened 2015-09-07
Fix tests

petemoore:


test/helper.js
```diff
@@ -33,7 +33,7 @@ if (!cfg.get('azure:accountKey') ||
 var webServer = null;
 mocha.before(async () => {
   webServer = await server('test');
-  webServer.setTimeout(500);
+  webServer.setTimeout(5000);
```
jonasfj: Explain this? This is webserver timeout of connections... killing them quickly means test process won't hang waiting for connections to close...

jonasfj:
`r-` pending comment being addressed on timeout issue.

Otherwise, this looks reasonable...

jonasfj:
Reopen when addressed.

test/helper.js
petemoore: The tests are consistently failing on travis due connection dropping. The tests started failing when this timeout was added. This was not noticed, partly because there was no badge to show that the CI was failing. Increasing this timeout fixes the tests.

See https://travis-ci.org/taskcluster/taskcluster-auth/builds. All of the failures are caused by this timeout.

jonasfj: hmm... Can we try with `1500` maybe that works better with nodes keepAlive setting...

PR #13, opened 2015-09-10
Used regexp again

jonasfj:
please check that you don't use the `pmoore2` client anymore... and then delete it... then merge this..


jonasfj:
Basically this broke our list-clients API... which broken our UI...

petemoore:
I saw this after I already applied the same change in b38ea26a7dbd061301fa92d1996b3023e18d6a74 - so it should be there already now. Closing...

PR #14, opened 2015-09-30
Support roles

jonasfj:
This is roles, these roles makes up rules, example:
```
A) assume:client-id:my-client -> [scope1]
B) assume:client-id:* -> [scope2]
```
(A) grants `scope1` to anyone with the scope `assume:client-id:my-client`.
(B) grans `scope2` to anyone with a scope that starts with `assume:client-id:`.

Note, clientIds are now identifiers humans can read, 1-22 chars `[a-z0-9_-]`. We no longer assign scopes to clients, instead a client gets the scope `assume:client-id:<clientId>`, so whenever we make a client, we also have to make a role for it. This is so that automated analysis is easier, as it doesn't have to learn about clients.

I also refactored a bit to get aws and azure specific parts out of `v1.js`


jonasfj:
Note, don't expect tests to work on travis... I'm seeing some issues with IAM... Intermittent so I think they are unrelated by I'll start digging into that.

djmitche:
Mixing the refactor in here was a bit mean :(

auth/data.js
```diff
@@ -1,206 +1,131 @@
...
-        client,
-        error,
-        reloadAt: now + options.cacheTimeout
-      };
-    }
-    if (entry.error) {
-      throw entry.error;
-    }
-    return entry.client;
+/** Get JSON representation of client */
```
djmitche: ..of a role..

jonasfj:
Yeah, sorry about the refactor... In many ways this almost a rewrite, so don't look too much at the diffs...

Main thing to look at is the ScopeResolver it's where the magic happens... Everything else is pretty much boilerplate, get/create/list/update entities, nothing special about the API logic..

auth/scoperesolver.js
```diff
@@ -0,0 +1,338 @@
...
+
+    // List of client objects on the form:
+    // {clientId, accessToken, expandedScopes: [...], updateLastUsed}
+    this._clients       = [];
+    // List of role objects on the form:
+    // {roleId: '...', scopes: [...], expandedScopes: [...]}
+    this._roles         = [];
+
+    // Mapping from clientId to client objects from _clients:
+    // {clientId, accessToken, expandedScopes: [...], updateLastUsed}
```
djmitche: `{clientId: client}`?

auth/scoperesolver.js
```diff
@@ -0,0 +1,338 @@
...
+var base        = require('taskcluster-base');
+var debug       = require('debug')('auth:ScopeResolver');
+
+class ScopeResolver extends events.EventEmitter {
+  /** Create ScopeResolver */
+  constructor() {
+    super();
+
+    // List of client objects on the form:
+    // {clientId, accessToken, expandedScopes: [...], updateLastUsed}
```
djmitche: maybe `updateLastUsedAt` - I assumed this was an async function to update the last-used field.

djmitche: OK that's not right either.  Maybe a comment about what the field means :)

auth/scoperesolver.js
```diff
@@ -0,0 +1,338 @@
...
+            scopes = _.union(scopes, ['assume:' + role.roleId]);
+          }
+          roles.push({roleId: role.roleId, scopes});
+        }
+      })
+    ]);
+
+
+    // Set _roles and _clients at the same time and immediately call
+    // _computeFixedPoint, so anyone using the cache is using a consistent one.
```
djmitche: I'm a little worried here that with multiple near-simultaneous updates, you may end up with a stale cache if an update other than the last takes the longest.

djmitche: In Python there are tools to serialize and debounce async functions like this, so that at least one invocation of the function *starts* after the last invocation.  Presumably similar things in JS?

auth/scoperesolver.js
```diff
@@ -0,0 +1,338 @@
...
+    this._roles = roles;
+    this._clients = clients;
+    this._computeFixedPoint();
+  }
+
+  /** Compute fixed point over this._roles, and construct _clientCache */
+  _computeFixedPoint() {
+    // Add initial value for expandedScopes for each role R
+    for (let role of this._roles) {
```
djmitche: nit: change `role` to `R` here for consistency with the comment

auth/scoperesolver.js
```diff
@@ -0,0 +1,338 @@
...
+    // allocations whenever possible.
+
+    // Rule (A) and (B) both requires the scope to start with "assume:"
+    if (scope.startsWith('assume:')) {
+      // A) We have scope = 'assume:<roleId>', so we can assume the role
+      if (scope.length === roleId.length + 7 && scope.endsWith(roleId)) {
+        return true;
+      }
+
+      // B) guard is on the form 'assume:<prefix>*' and we have a scope on the
```
djmitche: s/guard/role/g

auth/scoperesolver.js
```diff
@@ -0,0 +1,338 @@
...
+      //    form 'assume:<prefix>...'. This is special rule, assigning
+      //    special meaning to '*' when used at the end of a roleId.
+      if (roleId.endsWith('*') && scope.slice(7).startsWith(roleId.slice(0, -1))) {
+        return true;
+      }
+    }
+
+    // C) We have scope as '<prefix>*' and '<prefix>' is a prefix of guard, this
+    //    is similar to rule (A) relying on the normal scope satisfiability.
+    if (scope.endsWith('*') && roleId.startsWith(scope.slice(7, -1))) {
```
djmitche: I think the idea here is to allow scope `ass*` to match role `assume:foo`, and this works because `"ass*".slice(7, -1)` is `""`.  But so is `"queue:*".slice(7, -1)`, meaning the scope `"queue:*"` will grant all roles.  No es bueno!

You could write
```
if (scope.endsWith('*') && ("assume:" + roleId).startsWith(scope.slice(0, -1)))
```
for correctness, but it might be easier to just move this under the `if (scope.startsWith('assume:'))` above and lose expansion of scopes like "assu*" into all roles (which we'd never want anyway)

auth/scoperesolver.js
```diff
@@ -0,0 +1,338 @@
...
+    for (let R of this._roles) {
+      let isFixed = false;
+      while (!isFixed) {
+        isFixed = true; // assume we have fixed point
+        for (let role of this._roles) {
+          // if R can assume role, then we union the scope sets, as there is a
+          // finite number of strings here this will reach a fixed-point
+          let test = (scope) => ScopeResolver.grantsRole(scope, role.roleId);
+          if (R.expandedScopes.some(test)) {
+            let count = R.expandedScopes.length;
```
djmitche: why not check this length before and after the inner loop, rather than at every iteration?

auth/scoperesolver.js
```diff
@@ -0,0 +1,338 @@
...
+    // Compute fixed-point of roles.expandedScopes for each role R
+    for (let R of this._roles) {
+      let isFixed = false;
+      while (!isFixed) {
+        isFixed = true; // assume we have fixed point
+        for (let role of this._roles) {
+          // if R can assume role, then we union the scope sets, as there is a
+          // finite number of strings here this will reach a fixed-point
+          let test = (scope) => ScopeResolver.grantsRole(scope, role.roleId);
+          if (R.expandedScopes.some(test)) {
```
djmitche: If I'm not mistaken, this is equivalent to (but faster than) `util.scopeMatch([["assume:" + role.roleId]], R.expandedscopes)` (just verifying my understanding)

auth/scoperesolver.js
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    // Construct client cache
+    this._clientCache = {};
+    for (let client of this._clients) {
+      var scopes = this.resolve(['assume:client-id:' + client.clientId]);
+      client.scopes = scopes; // for createSignatureValidator compatibility
+      client.expandedScopes = scopes;
+      this._clientCache[client.clientId] = client;
+    }
```
djmitche: What if I have roles for `assume:client-id:*` or some longer prefix (the format of which would depend on how we name clientIds)?

auth/scoperesolver.js
```diff
@@ -0,0 +1,338 @@
...
+      // under-approximation
+      if (!scope.startsWith('assume:') && !scope.endsWith('*')) {
+        continue;
+      }
+
+      // For each role, expand if the role can be assumed, note we don't need to
+      // traverse the scopes added... As we the fixed-point for all roles.
+      for (let role of this._roles) {
+        if (ScopeResolver.grantsRole(scope, role.roleId)) {
+          scopes = _.union(scopes, role.scopes);
```
djmitche: Is it legit to modify the object while iterating over it?

djmitche: Oh, not modifying, never mind :)

auth/v1.js
```diff
@@ -0,0 +1,615 @@
...
+  route:      '/clients/:clientId',
+  name:       'createClient',
+  input:      'create-client-request.json#',
+  output:     'create-client-response.json#',
+  scopes:     [['auth:create-client:<clientId>']],
+  deferAuth:  true,
+  stability:  'stable',
+  title:      "Create Client",
+  description: [
+    "Words..."
```
djmitche: :)

auth/v1.js
```diff
@@ -0,0 +1,615 @@
...
+  route:      '/clients/:clientId/reset',
+  name:       'resetAccessToken',
+  input:      undefined,
+  output:     'create-client-response.json#',
+  scopes:     [['auth:reset-access-token:<clientId>']],
+  deferAuth:  true,
+  stability:  'stable',
+  title:      "Reset `accessToken`",
+  description: [
+    "Words..."
```
djmitche: Unless I misunderstand, this will immediately invalidate all temporary credentials issued with this clientId, too, right?  Worth noting in the docs.

jonasfj: yes, hmm... I should probably write docs :) rather than "words..." too...

test/grantsrole_test.js
```diff
@@ -0,0 +1,103 @@
...
+      result:   true
+    }, {
+      scope:    'assume:task-run-id:12345',
+      role:     'task-run-*',
+      result:   true
+    }, {
+      scope:    'assume:task-run-id:12345',
+      role:     'client-id:*',
+      result:   false
+    },
```
djmitche: I'm happy to see tests!!  Let's add some negative cases, too, especially as a regression test:
```
{
 scope: 'foo:*',
 role: "client-id:x",
 result: false
}
```

test/grantsrole_test.js
```diff
@@ -0,0 +1,103 @@
...
+  ].forEach(({scope, role, result}) => {
+    test(`grantsRole(${scope}, ${role}) === ${result}`, () => {
+      assert(ScopeResolver.grantsRole(scope, role) === result,
+             `Expected grantsRole(${scope}, ${role}) === ${result}`);;
+    });
+  });
+});
```
djmitche: Aww, no tests for `_computeFixedPoint`?

djmitche:
Aside from my usually "God's code goes right in; everyone else brings tests":

I'm worried about caching everything pre-expanded in memory, and refreshing the entire cache on every update.  For two reasons.  First, as the set of roles grows, this could begin to take a non-trivial amount of time, causing "jank" in the auth endpoint every time something's updated (since no requests will be handled while the fixed point computation is running).  Basically you're doing a lot of calculation that may never be used.  Second, this won't work at all once we have parameterized roles, since the parameters aren't available until the hawk authentication request arrives.

What do you think about refactoring this to only expand clientIds to their fixed point on demand?

auth/scoperesolver.js
jonasfj: good point, I'll serialize these... we rarely have many reloads, but no reason they shouldn't be serial...

Note: we do reload the cache every 20min, so stale cache is not a huge problem... This reload stuff is just best effort...

djmitche:
..and potentially also caching the expansion of any scopesets attached to hawk credentials.  Those could be cached based on a canonical representation of the scopeset -- maybe sorted and newline-separated (assuming newlines are actually forbidden in scopes)?

auth/scoperesolver.js
jonasfj: correct, I'll fix this, and implement it as two... 

We do want scope `a*` to grant all roles... you can't just break the semantics for `*` arbitrarily. It would be an absolute chaos, the end of civilization, zombies would rise and eat us all... I don't want to be blamed for that :)

auth/scoperesolver.js
jonasfj: good point... I previously did a breath first fixed point computation which it's a  left over from... I refactored to a  do-while...

auth/scoperesolver.js
jonasfj: correct, added comment to `grantsRole` doc-comment.

auth/scoperesolver.js
jonasfj: yes,

clientIds are now 1 to 22 chars `[a-z0-9_-]` so we can give human readable clientIds.

test/grantsrole_test.js
jonasfj: Agree... you found a pretty nasty case...

test/grantsrole_test.js
jonasfj: there is some in the `role_test.js` mainly done so because I prefer testing everything at once...

Otherwise it's a lot of code adding things entities, etc... The role_test also hits many of the reload points... 

But I probably need a test for recursive roles, normal and prefix roles are already covered...

djmitche: IMHO all of this is security sensitive and needs unit testing in addition to integration testing (which of necessity can only test a few permutations of inputs).

jonasfj:
> I'm worried about caching everything pre-expanded in memory, and refreshing the entire cache on every update.
We don't do that... We only reload the role or client that was modified. We don't reload everything (only every 20 min, could change that).

> causing "jank".... (since no requests will be handled while the fixed point computation is running)
False, the previously computed cache is available for resolving scopes while the cache is refreshing.. Changes to roles/clients are not instantaneous, they may in worst case take up to 20min, but because of the pulse exchanges they'll likely be up-to-date in 1-2s.

 > Basically you're doing a lot of calculation that may never be used.
Because of how scopes work we need to have all roles in memory. Precomputing the fixed-point should make this relatively fast though.

> What do you think about refactoring this to only expand clientIds to their fixed point on demand?
Only load the role when someone uses it. That will violate the `*` semantics.
And it would likely be slow...

> ..and potentially also caching the expansion of any scopesets attached to hawk credentials.
> Those could be cached based on a canonical representation of the scopeset
I considered it, but I suspect it's better to invest in being fast, storing things is the most expensive operation on a computer. And task specific credentials will probably get per task roles...  even if only used by the queue.

Anyways, I want to see how this performs before I invest too much in perf optimization, I might already have done too much premature optimization.

> maybe sorted and newline-separated (assuming newlines are actually forbidden in scopes)?
They aren't... it causes some problems that I have bugs for to fix. I don't think we'll restrict character sets allowed in scopes... But related issues will be fixed.

-----
Regarding size of things, I'm not too worried that we'll have roles and clients enough to consume 100 MB of memory. I'm pretty sure JS internalizes the strings and there is a finite number of those.
And I don't keep the role description in memory (which is probably a lot of text).

test/grantsrole_test.js
jonasfj: I'll add more tests... but because semantics are clean, there really isn't all that many permutations. There is with `grantsRole` which is why it has unit tests...

jonasfj:
@djmitche, so playing with this some more... It does get slow if I have 50 roles such that role1 -> role2, and role2 -> role3.... but that's a degenerate case.  And there is lots of room for optimizing the fixed-point computation... In general I'm not super worried as we won't be modifying these roles constantly.

It'll happen at most a few times a day. I find it unlikely that it'll slow down auth, but if it does we do have retries for everything.. And we can optimize it... 

My biggest fear computation wise is the `resolve()` function which is called for all scopes on all requests... it's a bit much... but we'll see :)
I might dive into some load-testing tomorrow... as I start on UI.

djmitche:
> False, the previously computed cache is available for resolving scopes while the cache is refreshing.. 

Correct me if I'm wrong, but since the JS is assynchronous and _computeFixedPoint operates synchronously, it blocks any other operation by that service during the cache refresh.  And since cache refreshes occur on receipt of a pulse message, they will happen more or less simultaneously across all dynos.  That will mean 1-2s of downtime for all taskcluster API calls (and potentially complications from the additional load as those API calls complete, are retried, etc.)

jonasfj:
> _computeFixedPoint operates synchronously
True... But the slow thing is usually the async load... if `_computeFixedPoint` starts to take 1-2s we should optimize it... there is certainly faster ways to compute it than what we currently do.

djmitche:
Yeah, this is O(n^2) though, and I'm worried that with the lack of parameterized roles n will get large quickly.  The original suggestion was to go to on-demand fixed-point generation now since, with the introduction of parameterized roles, we'll need it anyway.  Anyway, refactor now or refactor later isn't much difference, so I'll stop hammering the point :)

jonasfj:
I doubt the feasibility of parameterized roles... As well as the need for them... 
The semantics will definitely be interesting...  Anyway, I suspect they'll rely on the same fixed-point solution, except they'll have to close the parameterized scopes on demand... Well depending on how smart we get, he he :)

jonasfj:
btw, I just did DFA generation so we can do scope resolution in O(scopes.lenght) time, not implemented yet, but the hard part of it is done...

auth/v1.js
```diff
@@ -0,0 +1,679 @@
...
+    "  * Gain access to various services guarded by this API.",
+    "",
+    "### Clients",
+    "The authentication service manages _clients_, at a high-level each client",
+    "consists of a `clientId`, an `accessToken`, expiration and description.",
+    "The `clientId` and `accessToken` can be used for authentication when",
+    "calling TaskCluster APIs.",
+    "",
+    "Each client is assigned a single scope on the form:",
+    "`assume:client-id:<clientId>`, this scopes doesn't really do much on its",
```
djmitche: `"`assume:client-id:<clientId>`.  This scope`

djmitche:
:thumbsup: given the new tests.  Note the typo I just commented on.

PR #15, opened 2015-10-09
Optimizations for fixed point

jonasfj:
@djmitche, can you take a quick look at this...
It should be easy to r+ as tests covers all of it...


jonasfj:
I suspsect this broken... additional test suggests bugs... I'm investigating...

PR #16, opened 2015-10-09
Optimizations for fixed point

jonasfj:
@djmitche, from the load of unit test cases in rolelogic_test.js I hope it's safe to say that this works...

Anyways, please have a look, all tests pass, and I think the resolver code is a lot faster now... that was my main concern before...

Note: the fixed point computation, but the time spent matching scopes for every single request that is validated...

### DIFF
--- a/auth/dfa.js
+++ b/auth/dfa.js
@@ -4,61 +4,10 @@ var Promise     = require('promise');
 var ScopeResolver = require('./scoperesolver');
 
 /**
- * Sort roles by roleId, such that 'a*', comes right after 'a'
- *
- * Example: ['', 'a', 'a*', 'a(', 'aa', 'b'] is a list sorted as such.
- * Notice that the difference from a normal sorted list is only that '*'
- * comes before any other characters. Normally 'a(' would come before 'a*',
- * but not here.
- *
- * We do this such that we get a list looking a but like this:
- *  1. client-id:a
- *  2. client-id:try
- *  3. client-id:try*
- *  4. client-id:try-more
- *  5. client-id:z
- *
- * The cool thing is that when generating a DFA for this list all the possible
- * candidates (1-5) have the same path until we reach the 11th character.
- * At the 11th character the string diverge and any DFA constructed must
- * naturally have more than one state. However, for 11th character our DFA still
- * only needs 3 states, one representing (1), (2-4) and (5). But sorting the
- * list of roles, we represent these subsets elegantly and efficiently using
- * array offsets in the list of roles.
- *
- * More details on this later, for now just know that it makes DFA construct
- * both efficient and elegant (not to mention easy).
- */
-let sortRolesForDFAGeneration = (roles) => {
-  return roles.sort((a, b) => {
-    let n = a.roleId.length;
-    let m = b.roleId.length;
-    if (n === m && a.roleId.startsWith(b.roleId.slice(0, -1))) {
-      if (a.roleId[n - 1] === '*') {
-        return -1;
-      }
-      if (b.roleId[n - 1] === '*') {
-        return 1;
-      }
-    }
-    if (a.roleId < b.roleId) {
-      return -1;
-    }
-    return 1;
-  });
-};
-
-// Export sortRolesForDFAGeneration
-exports.sortRolesForDFAGeneration = sortRolesForDFAGeneration;
-
-/**
  * Compare scopes a and b to see which comes first if sorted
  * Such that 'a*' comes before 'a', but otherwise normal order.
  *
  * Example: ['*', '', 'a*', 'a', 'a(', 'aa', 'b'] is a list sorted as such.
- *
- * Notice that this is different from `sortRolesForDFAGeneration` as '*' comes
- * before the empty string.
  *
  * The reasoning for this sorting is pretty simple. If we have a set of scopes:
  *   ['a', 'a*', 'ab', 'b']
@@ -122,7 +71,8 @@ let scopeCompare = (a, b) => {
       if (a.startsWith(b)) {
         return -1;
       }
-    } else if (b[n - 1] === '*') {
+    }
+    if (m > n && b[m - 1] === '*') {
       if (b.startsWith(a)) {
         return 1;
       }
@@ -139,6 +89,7 @@ let sortScopesForMerge = (scopes) => {
 
 // Export sortScopesForMerge
 exports.sortScopesForMerge = sortScopesForMerge;
+
 
 /**
  * Take two sets of sorted scopes and merge them, normalizing in the process.
@@ -229,6 +180,36 @@ let mergeScopeSets = (scopes1, scopes2) => {
 
 // Export mergeScopeSets
 exports.mergeScopeSets = mergeScopeSets;
+
+/**
+ * Sort roles by roleId, such that 'a', comes right after 'a*'
+ *
+ * Example: ['', 'a*', 'a', 'a(', 'aa', 'b'] is a list sorted as such.
+ *
+ * We do this such that we get a list looking a but like this:
+ *  1. client-id:a
+ *  3. client-id:try*
+ *  2. client-id:try
+ *  4. client-id:try-more
+ *  5. client-id:z
+ *
+ * The cool thing is that when generating a DFA for this list all the possible
+ * candidates (1-5) have the same path until we reach the 11th character.
+ * At the 11th character the string diverge and any DFA constructed must
+ * naturally have more than one state. However, for 11th character our DFA still
+ * only needs 3 states, one representing (1), (2-4) and (5). But sorting the
+ * list of roles, we represent these subsets elegantly and efficiently using
+ * array offsets in the list of roles.
+ *
+ * More details on this later, for now just know that it makes DFA construct
+ * both efficient and elegant (not to mention easy).
+ */
+let sortRolesForDFAGeneration = (roles) => {
+  return roles.sort((a, b) => scopeCompare(a.roleId, b.roleId));
+};
+
+// Export sortRolesForDFAGeneration
+exports.sortRolesForDFAGeneration = sortRolesForDFAGeneration;
 
 /**
  * Build a DFA of states of the form:
@@ -342,6 +323,7 @@ let generateDFA = (roles, i, n, k, sets, implied) => {
     implied = sets.push([role, implied]) - 1;
     // And change the default transition for the current state to implied
     state.default = implied;
+    state.end = implied;
     // Now we move to the next role, if there is one
     j += 1;
     if (j >= n) {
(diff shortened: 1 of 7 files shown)


auth/dfa.js
```diff
@@ -0,0 +1,348 @@
...
+    }
+    return 1;
+  });
+};
+
+// Export sortRolesForDFAGeneration
+exports.sortRolesForDFAGeneration = sortRolesForDFAGeneration;
+
+/**
+ * Compare socpes a and b to see which comes first if sorted
```
djmitche: typo

auth/dfa.js
```diff
@@ -0,0 +1,348 @@
...
+  }
+
+  return state;
+};
+
+// Export generateDFA
+exports.generateDFA = generateDFA;
+
+/**
+ * Builds a resolver that given a scope returns a set of roles.
```
djmitche: Is that backward?

auth/dfa.js
```diff
@@ -0,0 +1,348 @@
...
+ *
+ * This also returns a lists of `sets`, where if you replace an entry in the
+ * list of sets it'll be the return value from the resolver.
```
djmitche: This doesn't make much sense.  Lists of sets of.. things? Return value from the resolver given what argument?

auth/dfa.js
```diff
@@ -0,0 +1,348 @@
+var _           = require('lodash');
+var assert      = require('assert');
+var Promise     = require('promise');
+var ScopeResolver = require('./scoperesolver');
+
+
+/** Sort roles by roleId, such that 'a*', comes right after 'a' */
```
djmitche: So basically EOF < `*` < all other characters..

auth/dfa.js
```diff
@@ -0,0 +1,348 @@
...
+    return 1;
+  });
+};
+
+// Export sortRolesForDFAGeneration
+exports.sortRolesForDFAGeneration = sortRolesForDFAGeneration;
+
+/**
+ * Compare socpes a and b to see which comes first if sorted
+ * Such that 'a*' comes before 'a', but otherwise normal order.
```
djmitche: `*` < EOF < all other characters

auth/dfa.js
```diff
@@ -0,0 +1,348 @@
...
+// Export mergeScopeSets
+exports.mergeScopeSets = mergeScopeSets;
+
+/**
+ * Build a DFA of states on the form:
+ * ```js
+ * {
+ *   'a': // next state if the current character is 'a'
+ *   'end': [] // set of roles granted if the scope ends here
+ *   'prefix': {...} // role granted if the scope matches so far
```
djmitche: This needs to be expanded out a bit.  It sounds like the states are nested, but what's the difference between "end" and "prefix", and why is the latter an object?

djmitche:
Sorry to basically only comment on the comments in your code -- I'm trying to look at this while multitasking and thus not concentrating very much.  I'm sure I could figure out what you're doing by reading the looking at the code more intently, but it's well-optimized and thus doesn't read well, and I suspect others may have a similarly hard time.

Can you write a text file or a longer comment describing the models you're applying here at a high level, so that it's easier to figure out where each function fits in?

jonasfj:
I'll add an explanation of how I convert to DFA... and reuse the same DFA after fixed-point computation...

As well as the reasoning behind the sort orders... it's a bit low-level for sure...

jonasfj:
@djmitche, not sure this enough... but take a look... some of the comments are a bit formal...
But I'ved tried to spice it up with some examples that it should be possible to follow...

Feel free to complain that it's not enough and I'll write some more :)

djmitche:
OK, I had a look at everything up through `buildResolver` so far.  I also made some comment changes - see https://github.com/taskcluster/taskcluster-auth/compare/optimizations-for-fixed-point...djmitche:optimizations-for-fixed-point -- the main thing being including `*` states in the DFA definition.

I've pretty much run out of time, so I'm going to brain-dump here a bit.  Sorry.  Also, kexec isn't npm installing, so I can't run the tests, so I couldn't verify some of these things.

The DFA as generated looks like this (I think -- I did it by hand):

 ```js
dfa=
 { 'a': { end: ['a'] },
   'b': { end: ['b'],
          prefix: 'b*',
          'c': { end: ['bc'] },
          '*': { end: ['b', 'bc'] } },
   'c': { end: ['c'] },
   '*': { end: [ 'a', 'b', 'b*', 'bc', 'c' ] } }
 ```

I think this would not recognize the scope `a*`, though -- the state `dfa['a']` is lacking a `*` transition.

Why did you choose to handle prefixes separately, rather than rolling them into the `end` for all nested states?  For example:

```js
dfa=
 { 'a': { end: ['a'] },
   'b': { end: ['b', 'b*'],
          'c': { end: ['b*', 'bc'] },
          '*': { end: ['b', 'b*', 'bc'] } },
   'c': { end: ['c'] },
   '*': { end: [ 'a', 'b', 'b*', 'bc', 'c' ] } }
 ```

(I just added `b*` to every end under `dfa.b`)

In the code generation, you create:

```js
case undefined:
  return sets[17];
...
case default:
  return sets[17];
```
which on the face of it suggests that the `case undefined` can just be omitted.  But I think there are cases where the default should not be returning that set -- specifically, if it contains explicit scopes rather than implicit scopes.  In the DFA above, scope `ax` would incorrectly return `['a']` just like scope `a`.

jonasfj:
> Why did you choose to handle prefixes separately, rather than rolling them into the end for all nested states? For example:
```js
dfa=
 { 'a': { end: ['a'] },
   'b': { end: ['b', 'b*'],
          'c': { end: ['b*', 'bc'] },
          '*': { end: ['b', 'b*', 'bc'] } },
   'c': { end: ['c'] },
   '*': { end: [ 'a', 'b', 'b*', 'bc', 'c' ] } }
```
(I just added b* to every end under dfa.b)

You added `b*` to every terminal under `dfa.b`. But there are more implicit states, basically any transition that doesn't exist leads to a final state that returns `[]`. So you would have to do:
```js
dfa=
 { 'a': { end: ['a'] },
   'b': { end: ['b', 'b*'],
          'a': { end: ['b*'] },
          'b': { end: ['b*'] },
          'c': { end: ['b*', 'bc'] },
          'd': { end: ['b*'] },
          'e': { end: ['b*'] },
          ...
          '*': { end: ['b', 'b*', 'bc'] } },
   'c': { end: ['c'] },
   '*': { end: [ 'a', 'b', 'b*', 'bc', 'c' ] } }
```
See above how we need a state for every letter in the alphabet, and all these dummy states aren't all that interesting. That's why I add the rule that if you traverse through a state with `prefix: 'b*'` then you've already got `b*`. It's sort of an non-terminal accepting state, in that you've got something, but you're not terminated yet and can continue matching :)

djmitche:
You're right, I meant to add a `default` case (which would correspond to `case default`):

```js
dfa=
 { 'a': { end: ['a'] },
   'b': { end: ['b', 'b*'],
          'c': { end: ['b*', 'bc'] },
          default: ['b*'],
          '*': { end: ['b', 'b*', 'bc'] } },
   'c': { end: ['c'] },
   default: [],
   '*': { end: [ 'a', 'b', 'b*', 'bc', 'c' ] } }
```

As it's written now, you're generating the DFA one way as data, then subtly (and I'm not sure correctly) modifying it when rendering the code for it.  My suggestion would be to do all of the logical transforms when generating the DFA, making the rendering much simpler.

jonasfj:
ah, default case might be interesting...

It certainly simplifies the `sets` thing, which is pretty nifty...But I think it would need some love to ensure that we don't explode the number of entries... which this would do, as we would have a `defaults` state on each state in the dfa for `['a*', 'aaaaaaaaaa']`. So number of entries in `sets` would be linear to the length of the scopes and not to the number of different terminal states..

There is something interesting here, will think about it...

jonasfj:
New `buildResolver` isn't integrated yet... but it's passing tests and generates a simpler DFA...
I'll add some docs for it later...

djmitche:
Have you addressed the correctness issues I raised?

jonasfj:
> Have you addressed the correctness issues I raised?

Correctness issues??? If you have any suggestions for things you don't think will pass unit tests... please suggest them. I love crazy test cases...

djmitche:
This part:

> I think this would not recognize the scope a*, though -- the state dfa['a'] is lacking a * transition

jonasfj:
yeah, I think that is addressed... Should also have worked before...
But by all means give a test case... `a*` grants roles `a` and `aa`... just added a test case to ease your mind...

Seriously though, if you have other test cases that worries you, please suggest them... it's pretty easy to define them in with the `testBuildResolver` function.

jonasfj:
@djmitche, have a look all tests pass... code is simplified, I think I updated most of the comments...
It's also a bit faster now... And with two fairly easy todos left in case we need more speed in the future :)

auth/dfa.js
```diff
@@ -0,0 +1,605 @@
...
+  roles = sortRolesForDFAGeneration(roles);
+  let sets = [[]];
+  let dfa = generateDFA(roles, 0, roles.length, 0, sets, 0);
+
+  // Render a DFA state to code
+  let renderDFA = (state, depth) => {
+    var d = '';
+    while (d.length < depth * 4) d += '    ';
+    var c = '';
+    if (typeof(state.end) === 'number') {
```
djmitche: I believe this must always be the case -- could this be replaced with an assertion on the typeof?

auth/dfa.js
```diff
@@ -0,0 +1,605 @@
...
+  //       this could probably be significantly faster. Might require some smart
+  //       cycle handling logic, but it seems fairly feasible.
+  //console.time("traveseImpliedRoles");
+  // For each role we compute the fixed-point the set of scopes the role expands
+  // to when all implied roles are considered. If roles are ordered
+  // unfortunately, ie. 1 -> 2 -> 3 -> 4, we can end up traversing the entire
+  // tree N times. Our unit tests is good example of this. To avoid these
+  // degenerate cases we'll quickly shuffle the roles around. In practice such
+  // degenerate cases probably don't exist, but processing order does affect
+  // performance, so we shuffle just for good measure.
```
djmitche: Why is this a win?  You're adding the cost of a shuffle to every operation, for the benefit of getting a better distribution of runtimes around the same mean.  Is there some O(1) trick you could use similar to finding a less-than-awful pivot for quicksort?

djmitche:
This looks good - my comments are very minor

auth/dfa.js
jonasfj: I think you're right, we always add `end` and nothing else `end == default`, but if in some future we break that variant and stop adding `end` when `end == default`, it'll be nice to have this condition.

But yeah, it's a nit.

auth/dfa.js
jonasfj: It only handles degenerate cases... we could probably do it smarter... but the whole search could be smarter...
Cost of doing the shuffle is nothing compared to the cost of exploring all the implied roles for each role.

jonasfj:
Okay, I'll try to roll this out today and see if it works :)

PR #19, opened 2015-10-23
add a validation test covering a schema with a scope

djmitche:


jonasfj:
Looks good... We should probably consider a negative test too...

jonasfj:
Lol, can't merge from phone because its failing... Sad when all of auth automation tests have been failing for months... I probably should fix that...

PR #20, opened 2015-10-26
add trailing `$` to scope pattern

djmitche:


jonasfj:
nice, thanks...

jonasfj:
@djmitche,
hmm... this is still wrong... can't deploy it... we have to do `"^[\\x20-\\x7e]*$"`

Just noticed after merging...

djmitche:
even in the yml?  but the validation tests passed :( :(

jonasfj:
Hmm, good question: http://yaml.org/spec/1.1/#id872840
Probably only if we use double quotes..

Just tried it with an online yaml to json thing... If we use double it is a problem otherwise not... Wow.. But hey, I get why most thing support unicode escape sequences..

Hmm, I wonder if js regexp support is good enough that it does matter... Hmm, after all \x20 becomes space... :)

PR #21, opened 2015-10-26
More scope tests

jonasfj:
Builds on top of #16  and just adds a few simple tests for scopes... Also requires scopes in roles to be unique... there is no reason to allow duplicates...


jonasfj:
Feel free to deal with #16 first, then this is easier to review :)

djmitche:
:thumbsup:

PR #23, opened 2015-10-30
Test for lastused and fix...

jonasfj:
More tests... cleanups and more tests... and fix for lastUsed not working...

@djmitche I think you said you filed a bug for this but I can't find it...

Note this won't roll out until the gaia patches have been migrated to all branches, as I'm not sure if scope changes could break things... So we don't roll... I think scope changes could break something... As long as task.scopes are submitted with `\n`, so we need to wait for someone in gaia to propagate this...


auth/scoperesolver.js
```diff
@@ -219,10 +230,11 @@ class ScopeResolver extends events.EventEmitter {
 
   /** Update lastDateUsed for a clientId */
   async _updateLastUsed(clientId) {
-    let client = await this._Client({clientId});
+    let client = await this._Client.load({clientId});
```
djmitche: It looks like this was the actual fix?

djmitche:
This looks good and I'm glad to see the tests.

auth/scoperesolver.js
jonasfj: yeah... really stupid bug... I suspect as a result the auth server having been crashing every 6 hours or so... interesting :)

PR #24, opened 2015-11-03
Bug 1221121: check input scopes on createRole

djmitche:


auth/v1.js
```diff
@@ -427,15 +427,19 @@ api.declare({
...
+    "If there already exists a role with the same `roleId`this operation",
+    "will fail. Use `updateRole` to modify an existing role.",
   ].join('\n')
 }, async function(req, res) {
   let roleId    = req.params.roleId;
   let input     = req.body;
 
   // Check scopes
-  if (!req.satisfies({roleId})) {
+  if (!req.satisfies({roleId}) !! !req.satisfies([input.scopes]) {
```
petemoore: should `!!` be `||`?

djmitche: oh, geez, yes

petemoore: haha, well no worries, we should probably make getting the CI tests working in travis a priority so these things would get picked up.

jonasfj:
nice catch... thanks...

jonasfj:
hmm... This broke my tests... he he... fixed and landing...

PR #25, opened 2015-11-03
Bug 1137740: limit temporary credentials, use client role

djmitche:
This limits the creds' lifetime to 3 days, and uses the `client-id:..`
role explicitly so that as the clientId's scopes change, so do the
temporary credentials.

### DIFF
--- a/auth/persona.js
+++ b/auth/persona.js
@@ -29,8 +29,11 @@ var renderIndex = async function(authFailed, req, res) {
     });
     credentials = taskcluster.createTemporaryCredentials({
       start:        new Date(),
-      expiry:       new Date(new Date().getTime() + 31 * 24 * 60 * 60 * 1000),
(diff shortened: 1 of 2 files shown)


jonasfj:
okay, will roll this out in a bit...

auth/persona.js
```diff
@@ -29,8 +29,11 @@ var renderIndex = async function(authFailed, req, res) {
     });
     credentials = taskcluster.createTemporaryCredentials({
       start:        new Date(),
-      expiry:       new Date(new Date().getTime() + 31 * 24 * 60 * 60 * 1000),
-      scopes:       client.expandedScopes(),
+      // good for three days
+      expiry:       new Date(new Date().getTime() + 3 * 24 * 60 * 60),
```
jonasfj: Oh, you're missing `* 1000` it's ms

PR #26, opened 2015-11-04
Allow all printable clientids

jonasfj:
https://bugzilla.mozilla.org/show_bug.cgi?id=1221242


djmitche:
Should we spell out this requirement somewhere in the docs, too?

jonasfj:
@djmitche, good point we should...

PR #27, opened 2015-11-16
Use login

jonasfj:
Remove all UI and redirect to `login.taskcluster.net`

Tests pass... we need to test this in production... to see if the redirect works...

I've already moved `tools.taskcluster.net` to use login... so this is just for backwards compatibility...


jonasfj:
re tests don't worry I have a port to new config and loader in progress... with that... it should be a bit easier to fix the travis tests.

djmitche:
This looks fine

Sorry for the delay -- I opened it up, saw "filed changed: 21" and tossed it into TaskWarrior.  Turns out it only took 20 seconds to review :)

PR #29, opened 2015-12-01
Added pulse credentials to encrypted test config file to get tests running

petemoore:
The tests are now running again. Currently only 3 are failing (and 271 are passing).

I also set travis to only allow one concurrent job, so now the failures are not because tests are running in parallel. Maybe these are genuine issues or faulty tests?

```
  1) api (roles) importClients (clean-up w. deleteClient):
     Error: socket hang up
      at createHangUpError (http.js:1473:15)
      at Socket.socketOnEnd (http.js:1569:23)
      at _stream_readable.js:944:16

  2) api (role logic) test with depth = 5 x 50:
     Error: timeout of 300000ms exceeded. Ensure the done() callback is being called in this test.
  
  3) api (role logic) assume* scope is *:
     Error: client with same clientId already exists, possiblyan issue with retry logic or idempotency
      at node_modules/taskcluster-client/lib/client.js:320:21
      at tryCallOne (node_modules/promise/lib/core.js:37:12)
      at node_modules/promise/lib/core.js:103:15
      at flush (node_modules/promise/node_modules/asap/raw.js:50:29)
```


petemoore:
Basically, I'm proposing we merge this PR and then raise a separate one to fix the broken tests above. Currently the tests are not running due to missing pulse credentials, so this is why I added the pulse credentials in this PR - so then we'll go from having no tests running to 271/274 tests passing...

P.S. to update the encrypted config file, I created a dummy branch with my public gpg key checked in, updated travis script to encrypt the README.md using my key and printing the armor'd result, and once that was working changed it to print the armor'd gpg encrypted version of the unencrpypted config file (using README.md was just to reduce risk of making a mistake on the way)! I'm hoping to find a solution to our secret sharing so we can avoid workarounds like this, but was cool that it worked. =)

jonasfj:
Honestly, I wanted this ported the new config module... which would make this easier... I started it but never finished that port...

PR #31, opened 2016-01-22
Testing fixes

djmitche:
This is a bunch of minor refactors to use newer `taskcluster-lib-*`, with the end result that it's a lot easier and faster to run the tests.

### DIFF
--- a/auth/data.js
+++ b/auth/data.js
@@ -1,19 +1,19 @@
 var debug       = require('debug')('auth:data');
-var base        = require('taskcluster-base');
+var Entity      = require('azure-entities');
 var assert      = require('assert');
 var _           = require('lodash');
 var taskcluster = require('taskcluster-client');
 
-var Client = base.Entity.configure({
+var Client = Entity.configure({
   version:          1,
-  partitionKey:     base.Entity.keys.StringKey('clientId'),
-  rowKey:           base.Entity.keys.ConstantKey('client'),
+  partitionKey:     Entity.keys.StringKey('clientId'),
+  rowKey:           Entity.keys.ConstantKey('client'),
   signEntities:     true,
   properties: {
-    clientId:       base.Entity.types.String,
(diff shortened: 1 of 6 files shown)


bin/server.js
```diff
@@ -10,189 +10,210 @@ var exchanges     = require('../auth/exchanges');
...
-      validator = await base.validator({
+var loader        = require('taskcluster-lib-loader');
+var app           = require('taskcluster-lib-app');
+
+// Create component loader
+let load = loader({
+  cfg: {
+    requires: ['profile'],
+    setup: ({profile}) => {
+      return base.config({
```
jonasfj: I know it's a pain, but upgrading base and using new config is also worth it...

Lot's of good things have been happening in API, from query string support for things like continuation tokens... to things like better error handlings and error messages...

auth/data.js
```diff
@@ -22,7 +22,7 @@ var Client = base.Entity.configure({
      * - lastRotated      // Last time accessToken was reset
      * (more properties may be added in the future)
      */
-    details:        base.Entity.types.JSON
+    details:        Entity.types.JSON
   },
   context:          ['resolver']
```
jonasfj: We could probably get rid of this circlar dependency... and take it as a parameter for `.json`

bin/server.js
djmitche: Sure, but it's not necessary here.  Well, in a subsequent patch, query string support might be useful (for listClients), but not in this PR.

auth/data.js
jonasfj: with `loader` it's pretty obvious that circular dependencies are bad.

jonasfj:
`r+`

My comment on fixing the circular dependency shouldn't block anything... it's just a nice to have.. .that'll probably make us happier further down the line... (also it would simplify tests)

PR #32, opened 2016-01-25
Bug 1242714: reject expired clients

djmitche:
This merges #9, then fixes the implementation to pass those tests.

It also includes all of the test refactoring from #31.


jonasfj:
`r+` seems solid to me...

PR #34, opened 2016-01-26
New client properties

djmitche:
Per https://gist.github.com/djmitche/f59a481e0c2580d6cdfb this adds client.scopes (leaving the expansion of `client-id:xxx` roles) and client.disabled, plus the API changes to make those work.

It also refactors `client` and `listClients` to pull their data from the resolver cache, rather than from Azure.  This will be handy when `listClients` grows the ability to filter by prefix.

Each commit is fairly self-contained, and tests pass after each one.

### DIFF
--- a/auth/data.js
+++ b/auth/data.js
@@ -83,7 +83,7 @@ Client.ensureRootClient = function(accessToken) {
   // Create client resolving conflicts by overwriting
   return Client.create({
     clientId:         'root',
-    description:      "Automatically created `root` client " +
+    description:      "Automatically created `root` client with star scopes " +
                       "for bootstrapping API access",
     accessToken:      accessToken,
     expires:          taskcluster.fromNow('24 hours'),
@@ -93,7 +93,7 @@ Client.ensureRootClient = function(accessToken) {
       lastDateUsed:   new Date().toJSON(),
       lastRotated:    new Date().toJSON()
     },
-    scopes:           [],
+    scopes:           ['*'],
     disabled:         0,
   }, true);
 };
@@ -135,29 +135,6 @@ Role.prototype.json = function() {
     scopes:         this.scopes,
     expandedScopes: this.resolver.resolve(scopes),
   };
-};
-
-/**
- * Ensure the role: client-id:root -> ['*'] exists
- *
- * Should only be called if the app is configured with a rootAccessToken.
- * Otherwise, app should assume whatever is in the table storage is the
- * root access token, and that appropriate role is attached.
- *
- * Basically, this is for bootstrapping only.
- */
-Role.ensureRootRole = function() {
-  let Role = this;
-  return Role.create({
-    roleId:       'client-id:root',
-    description:  "Automatically created role for bootstrapping the `root` "+
-                  "client.",
-    scopes:       ['*'],
-    details: {
-      created:        new Date().toJSON(),
-      lastModified:   new Date().toJSON()
-    }
-  }, true);
 };
 
 // Export Role
--- a/auth/data.js
+++ b/auth/data.js
@@ -14,6 +14,17 @@ var Client = Entity.configure({
     description:    Entity.types.Text,
     accessToken:    Entity.types.EncryptedText,
     expires:        Entity.types.Date,
+    details:        Entity.types.JSON
+  },
+  context:          ['resolver']
+}).configure({
+  version:          2,
+  signEntities:     true,
+  properties: {
+    clientId:       Entity.types.String,
+    description:    Entity.types.Text,
+    accessToken:    Entity.types.EncryptedText,
+    expires:        Entity.types.Date,
     /**
      * Details object with properties:
      * - created          // Time when client was created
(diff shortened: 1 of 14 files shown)


auth/data.js
```diff
@@ -22,14 +33,24 @@ var Client = Entity.configure({
      * - lastRotated      // Last time accessToken was reset
      * (more properties may be added in the future)
      */
-    details:        Entity.types.JSON
+    details:        Entity.types.JSON,
+    scopes:         Entity.types.JSON,  // new in v2
+    disabled:       Entity.types.Number // new in v2
```
jonasfj: `JSON` is probably the best way to serialize a boolean... unless of course you want to be able to use it in a query...

Anyways, sad we don't have boolean in `azure-entities`, it's kind of bad to miss that :)

auth/v1.js
```diff
@@ -134,14 +128,14 @@ api.declare({
 }, async function(req, res) {
   let clientId = req.params.clientId;
 
-  // Load client
-  let client = await this.Client.load({clientId}, true);
+  // Load client from the resolver's cache
+  let client = this.resolver.jsonClient(clientId);
```
jonasfj: I have mixed feeling about loading the clients from memory.. notably, they could be slightly out of date as it takes a few ms for the pulse message purging the client to arrive... And a few ms to reload it...

So we now have race conditions in the editor that could result in someone seeing and old client.. IMO the purge of clients from memory based on pulse message is sort of a best effort service..

There is a reason we purge all clients every 20 minutes, that because then we don't have to worry about loosing a push message. In this case it would definitely feel safer if we loaded from azure...

It's not like the cost of loading for this purpose is significant. The cost of having out-of-date information is however a lot higher... Out of date authentication by 20min is not ideal, but certainly more acceptable.

jonasfj: Same goes when listing clients. The cost of the extra load is small, let's just do it.

auth/v1.js
```diff
@@ -167,13 +161,16 @@ api.declare({
...
+    "The caller's scopes must satisfy `scopes`."
   ].join('\n')
 }, async function(req, res) {
   let clientId  = req.params.clientId;
   let input     = req.body;
+  let scopes    = input.scopes || [];
 
   // Check scopes
-  if (!req.satisfies({clientId})) {
+  if (!req.satisfies({clientId}) || !req.satisfies([scopes])) {
```
jonasfj: nice catch :)

auth/v1.js
```diff
@@ -315,11 +316,112 @@ api.declare({
     return res.status(404).json({message: "Client not found!"});
   }
 
-  // Reset accessToken
+  let added = _.without.apply(_, [input.scopes].concat(client.scopes));
+  if (!req.satisfies([added])) {
```
jonasfj: btw, solution outlined in the bug you NI'ed me on was pretty awesome go with that :)

auth/v1.js
```diff
@@ -315,11 +316,112 @@ api.declare({
...
+
+
+/** Disable client */
+api.declare({
+  method:     'post',
+  route:      '/clients/:clientId/disable',
+  name:       'disableClient',
+  input:      undefined,
+  output:     'get-client-response.json#',
+  scopes:     [['auth:enable-client:<clientId>']],
```
jonasfj: different scope maybe?

auth/v1.js
```diff
@@ -315,11 +316,112 @@ api.declare({
...
+  deferAuth:  true,
+  stability:  'stable',
+  title:      "Enable Client",
```
jonasfj: And different title?

jonasfj: Copy/paste rules I know..

auth/v1.js
```diff
@@ -637,7 +739,9 @@ api.declare({
         lastModified: new Date().toJSON(),
         lastDateUsed: new Date().toJSON(),
         lastRotated:  new Date().toJSON()
-      }
+      },
+      scopes:         [],
+      disabled:       0
```
jonasfj: I think it's safer to just delete this API end-point... we're not going to be using it anyways..

It was for legacy import, and we don't have export anymore...

jonasfj: (optional, since it's not related to this PR)

auth/v1.js
jonasfj: (optional, since it's not related to this PR)

jonasfj:
A few small things, but everything looks good... 

The scope for disable, and possibly rethink whether returning clients from memory is smart... If they are just slightly out of date it's very annoying when editing...

Note: technically, our `updateClient` doesn't work very well if two people edit it at the same time. So using an out-of-date client representation would be pretty bad.

auth/v1.js
djmitche: Hm, last we talked design, we agreed it was safe to do prefix filtering for clients because we weren't doing table scans.  With 1000's of clients, are you still comfortable with the extra load?

auth/v1.js
djmitche: This method didn't take scopes before, so nothing's caught :)

auth/v1.js
djmitche: Yeah, I'll loop back and do both API methods at once.  There's a skipped test that will go green, too.

auth/v1.js
djmitche: Oh I relish the opportunity to delete code :)

djmitche:
How would you feel about using cached data for `listClients` and data straight from Azure for `client`?  Good compromise?

auth/v1.js
jonasfj: Hmm... Honestly, how bad can it be... If we do continuationtoken in the future loading shouldn't be a problem...

Maybe one day we'll fix azure-entities to support prefix matching...I had it working in a fork, but it requires full table migration as we change the keys...

BTW, usually the query operation on azure returns upwards 1k entries... Per request... So even 10k users should be okay... Especially if we do continuationtoken at some point..

But let's first get that 1k users :)

jonasfj: And yeah, I realize we talked about doing prefix matching because we had all clients in memory... How naive are we :)

He he, actually both might work, but slow and safe might be best... Otherwise the compromise is also acceptable...

djmitche: OK, I'll split out this refactor (it was a little fiddly to get right) into a separate PR and leave it there so it's handy if we find we want it.

djmitche:
OK, I think that covers all of the comments -- anything else before I merge?

jonasfj:
Ship it! :)

Looks really great!

djmitche:
This is deployed now, with no ill effects

PR #35, opened 2016-01-26
[DO NOT MERGE] Use resolver cache

djmitche:
This is an offshoot of #34 that switches `client` and `listClients` to read from the resolver's cache rather than from the Azure tables.  In that PR, we decided not to do this yet, as it introduces the possibility of serving stale data from these requests.


gregarndt:
This has been set as do not merge since it was opened and not updated in over a year.  Can we close this PR?

djmitche:
Yeah, thanks.

PR #36, opened 2016-01-27
Upgrade node4

jonasfj:
Warning this needs redefinition of heroku config:

All env vars from config.yml and `NODE_ENV`


jonasfj:
Looks like we'll have to fiddle a bit more with travis...

PR #37, opened 2016-01-29
handle missing config a little better

djmitche:
With this, you can omit `influx` and `azure` from user-config.yml and still have things run.


jonasfj:
Looks good to me... but why didn't you have to break travis :)

Ah, you're pushing to your own repo...

@djmitche generally just push to a branch on taskcluster repo, and delete the branch when the PR is merged... then travis will with with credentials...

PR #38, opened 2016-01-29
Util methods

djmitche:
Got another one for you, @jonasfj

### DIFF
--- a/auth/scoperesolver.js
+++ b/auth/scoperesolver.js
@@ -252,11 +252,14 @@ class ScopeResolver extends events.EventEmitter {
   }
 
   /**
-   * Return set of scopes that `scopes` can be expanded to when assuming all
-   * authorized roles.
+   * Return a normalized set of scopes that `scopes` can be expanded to when
+   * assuming all authorized roles.
    */
+
   resolve(scopes) {
-    let granted = dfa.sortScopesForMerge(_.clone(scopes));
+    // use mergeScopeSets to eliminate any redundant scopes in the input (which will
+    // cause redundant scopes in the output)
+    let granted = dfa.mergeScopeSets(dfa.sortScopesForMerge(_.clone(scopes)), []);
     for (let scope of scopes) {
       let found = this._resolver(scope);
       if (found.length > 0) {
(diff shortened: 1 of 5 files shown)


auth/scoperesolver.js
```diff
@@ -252,11 +252,12 @@ class ScopeResolver extends events.EventEmitter {
...
   /**
-   * Return set of scopes that `scopes` can be expanded to when assuming all
-   * authorized roles.
+   * Return a normalized set of scopes that `scopes` can be expanded to when
+   * assuming all authorized roles.
    */
+
   resolve(scopes) {
-    let granted = dfa.sortScopesForMerge(_.clone(scopes));
+    let granted = dfa.sortScopesForMerge(ScopeResolver.normalizeScopes(scopes));
```
jonasfj: This is hot code... if we want to do this at the very least leave a `//TODO: Avoid using ScopeResolver.normalizeScopes and normalize scopes using sorted result`

Basically, `ScopeResolver.normalizeScopes` is O(n^2) but after the sort by `dfa.sortScopesForMerge` normalization is trivially done in O(n), see:
https://github.com/taskcluster/taskcluster-auth/blob/3c3b0c9f9a98baa23839ee0d8b0accff878f462f/auth/dfa.js#L152-L166

Technically, this could be done with `mergeScopeSets(dfa.sortScopesForMerge(_.clone(scopes)), [])` which probably isn't even that bad an idea...

jonasfj: I'm saying this because we have spent time making sure this code path was fast.. :)

auth/v1.js
```diff
@@ -688,6 +698,41 @@ api.declare({
   return res.reply();
 });
 
+/** Expand a scopeset */
+api.declare({
+  method:     'delete',
```
jonasfj: he he, interesting choice of method... May I suggest `get`.. 

You have same copy/pasta issue below :)

jonasfj:
`r+` But please do fix the `delete` to `get` change :)

djmitche:
Updated -- thanks!

PR #39, opened 2016-02-03
move local signature validator from tc-lib-api

djmitche:


package.json
```diff
@@ -1,42 +1,46 @@
...
-    "body-parser":                      "^1.12.4",
-    "taskcluster-client":               "^0.23.9",
-    "taskcluster-lib-loader":           "^0.1.1",
-    "taskcluster-lib-app":              "^0.8.10",
-    "express":                          "^4.12.4",
-    "babel":                            "^5.8.23",
-    "kexec":                            "^1.2.0",
-    "fast-azure-storage":               "^0.3.2",
-    "hawk":                             "2.3.0"
+    "aws-sdk": "^2.2.0",
```
jonasfj: But what about all my awesome indentation :(

(don't worry I won't fight you... I know I've lost @jhford won this battle over me a while ago)

jonasfj:
`r+`

package.json
djmitche: I just used `npm install --save`

djmitche:
OK, I added a bunch of tests, based loosely on tc-lib-api:test/auth_test.js, but omitting the api-specific bits and testing a lot more of the certificate-evaluation bits.  I found two (very minor) bugs in certificate evaluation!

djmitche:
BTW, with this in place I'll be a lot more comfortable working on the named-temporary-credentials stuff, as I can more reliably test it and make sure I haven't broken verification of other kinds of signatures.

PR #40, opened 2016-02-04
Test end point

jonasfj:
Just wanted to outline this idea...

No idea when I'll have time to patch `tc-lib-api` to support this... or write tests...
Also this might need `req.clientId()` to be added too...

Also it would take a while to port all the client tests to use this...
But I suspect an end-point like this would be great for testing client implementations.

Aki, started work on a python 3.5 client in: https://bugzilla.mozilla.org/show_bug.cgi?id=1245835

If we can deploy this, we can use him as guinea pig to see if writing test that way makes sense.
I suspect this end-point should be enough to write solid tests for authentication, temporary credentials (with and without scope restrictions), as well as `authorizedScopes`.

---
Note: this is blocked on moving of signature validator and can't really move forward without that.


jonasfj:
@djmitche I think this is pretty decent now... Have a look and let me know...

djmitche:
You should probably rebase this on something other than my "WIP" commit  0298aa1 :)

auth/v1.js
```diff
@@ -762,6 +763,57 @@ api.declare({
...
+  output:     'test-authenticate-response.json#',
+  stability:  'experimental',
+  title:      "Test Authentication",
+  description: [
+    "Utility method to test your implementation of the authentication scheme.",
+    "The request payload carries a set of scopes that the client has",
+    "and a set of scopes that the request is required to have.",
+    "",
+    "You can authenticate against this end-point with `tester` as `clientId`",
+    "and `no-secret` as `accessToken`.",
```
djmitche: Rephrase this a little, since I didn't understand until I read the code:
---
Utility method to test client implementations of TaskCluster authentication.

Rather than using real credentials, this endpoint accepts requests with clientId `tester` and accessToken `no-secret`.  That client's scopes are based on `clientScopes` in the request body.

auth/v1.js
```diff
@@ -762,6 +763,57 @@ api.declare({
 });
 
 
+api.declare({
+  method:     'post',
+  route:      '/test-authenticate',
```
djmitche: It'd be good to have a unit test or two for this method.

jonasfj: See `testauth_test.js`

auth/v1.js
jonasfj: good suggestion...

jonasfj:
merged in master... Not sure why I had a weird base... Probably because I was bisecting it earlier...

djmitche:
actually I don't see this merged..

PR #41, opened 2016-02-08
Support named temporary credentials

djmitche:
Credentials are {id, key, cert}.  A "simple" temporary credential has
the issing client as its `id`, scopes and expiry in the cert, all signed
to generate a new key.  "Named" temporary credentials further include a
name and an issuer in the certificate; the issuer is the clientId that
created the credential (and the corresponding accessToken is used in the
signature), while the name is the clientId that should be supplied as
`id`.  The result is that temporary credentials can have a meaningful
id that is useful for logging and for identifying the credentials.

### DIFF
--- a/auth/signaturevalidator.js
+++ b/auth/signaturevalidator.js
@@ -184,8 +184,10 @@ var limitClientWithExt = function(credentialName, issuingClientId, accessToken, 
  * The function returned takes an object:
  *     {method, resource, host, port, authorization}
  * And returns promise for an object on one of the forms:
- *     {status: 'auth-failed', message}, or
- *     {status: 'auth-success', clientId, scheme, scopes}
+ *     {status: 'auth-failed', message},
+ *     {status: 'auth-success', clientId, scheme, scopes}, or
+ *     {status: 'auth-success', clientId, scheme, scopes, hash}
+ * where `hash` is the payload hash.
  *
  * The `expandScopes` applies any rules that expands scopes, such as roles.
  * It is assumed that clients from `clientLoader` are returned with scopes
--- a/auth/signaturevalidator.js
+++ b/auth/signaturevalidator.js
@@ -87,15 +87,6 @@ var limitClientWithExt = function(credentialName, issuingClientId, accessToken, 
 
     // Check clientId validity
     if (issuingClientId !== credentialName) {
-      if (!cert.clientId) {
-        throw new Error('ext.certificate.clientId must be set when ext.certificate.issuer is given');
-      }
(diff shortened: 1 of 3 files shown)


djmitche:
Hm, hold off, tests aren't passing.

djmitche:
OK, I forgot to adjust the response schema.  Tests pass now.
```
  347 passing (2m)
  1 pending
```

djmitche:
renamed name to clientId.

auth/signaturevalidator.js
```diff
@@ -71,23 +85,48 @@ var limitClientWithExt = function(client, ext, expandScopes) {
...
-        'version:'  + '1',
-        'seed:'     + cert.seed,
-        'start:'    + cert.start,
-        'expiry:'   + cert.expiry,
-        'scopes:',
-      ].concat(cert.scopes).join('\n'))
+    var sigContent = []
+    sigContent.push('version:'    + '1');
+    if (cert.clientId && cert.issuer) {
+      sigContent.push('clientId:' + cert.clientId);
```
jonasfj: Why would certificate include `clientId`? Let's not duplicate...

Just use `credentialName` instead of `cert.clientId`

auth/signaturevalidator.js
```diff
@@ -71,23 +85,48 @@ var limitClientWithExt = function(client, ext, expandScopes) {
...
+      sigContent.push('issuer:'   + cert.issuer);
```
jonasfj: This not necessary, not that I care to object... :)

it just seems wrong to put `issuer` under the signature only when `issuer != clientId` I see no need for this...

auth/signaturevalidator.js
djmitche: Hah, so that's better than "name"?

auth/signaturevalidator.js
djmitche: I see it as a worse sin to include information in a "certificate" that is not covered by that certificate's signature.

jonasfj: the signature isn't covered... but okay, you can cover it... nor harm...

auth/signaturevalidator.js
jonasfj: no I'm saying `cert.clientId` should not exist

jonasfj: the line should read `sigContent.push('clientId:' + credentialName);`

djmitche:
OK, updated to omit clientId from the certificate.  Without objection I'll land this tomorrow.

jonasfj:
`r+`

PR #42, opened 2016-02-10
add check-staging script

djmitche:
This provides a quick sanity check that the staging auth site is working.


jonasfj:
`r+` but don't run on production tables... in case we upgrade something it's really bad...

djmitche:
https://bugzilla.mozilla.org/show_bug.cgi?id=1247425 for that

PR #43, opened 2016-02-10
add test for signaturevalidator to return a hash

djmitche:
This wouldn't have caught the failure (since it doesn't test the JSON validation), but it's worth doing.


jonasfj:
Looks okay... we should probably do tests using hawk... but this is a solid start...

`r+`

djmitche:
This is using hawk..

PR #44, opened 2016-02-11
Better explanatory message when authorized scopes are not satisfied.

walac:
When the clientId doesn't satisfies the authorized scopes, report the
client scopes and the authorized scopes in the error messsage.


djmitche:
Looks good.  While we're on these messages, can we use the "satisifies" terminology instead?  "Supplied credentials do not satisfy authorizedScopes; credentials have scopes ${scopes}; authorizedScopes are ${authorizedScopes}"

walac:
> Looks good. While we're on these messages, can we use the "satisifies" terminology instead? "Supplied credentials do not satisfy authorizedScopes; credentials have scopes ${scopes}; authorizedScopes are ${authorizedScopes}"

Ok, fixed.

djmitche:
:+1:

PR #46, opened 2016-02-12
Add sentry integration

imbstack:
Should do what it says on the tin. Once this is deployed, I'll add the sentry dsn to heroku and we'll get exceptions reported to sentry.


djmitche:
Looks good.  Don't forget to update the config in both staging and production.

PR #47, opened 2016-02-12
Add testAuthenticationGet

djmitche:
A new method for testing bewits, which only work for GET requests.  Plus tests.

### DIFF
--- a/auth/v1.js
+++ b/auth/v1.js
@@ -816,6 +816,58 @@ api.declare({
   });
 });
 
+api.declare({
+  method:     'get',
+  route:      '/test-authenticate-get/',
+  name:       'testAuthenticateGet',
+  output:     'test-authenticate-response.json#',
+  stability:  'experimental',
+  title:      "Test Authentication (GET)",
+  description: [
+    "Utility method similar to `testAuthenticate`, but with the GET method,",
+    "so it can be used with signed URLs (bewits).",
+    "",
+    "Rather than using real credentials, this endpoint accepts requests with",
+    "clientId `tester` and accessToken `no-secret`. That client's scopes are",
+    "`['test:*', 'auth:create-client:test:*']`.  The call fails if the ",
+    "`test:authenticate-get` scope is not available.",
+    "",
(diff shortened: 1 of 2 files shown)


djmitche:
I just updated this to take a parameter instead, since to test named temporary credentials, I will need to have the clientId associated with `auth:create-client:<clientId>`.

auth/v1.js
```diff
@@ -816,6 +816,55 @@ api.declare({
...
+  stability:  'experimental',
+  title:      "Test Authentication (GET)",
+  description: [
+    "Utility method similar to `testAuthenticate`, but with the GET method,",
+    "so it can be used with signed URLs (bewits).",
+    "",
+    "Rather than using real credentials, this endpoint accepts requests with",
+    "clientId `tester` and accessToken `no-secret`. That client's scopes are",
+    "`['test-authenticate-get', clientScope]`.  The call fails if the",
+    "`test-authenticate-get` scope is not avaialble (e.g,. via authenticateScopes)",
```
jonasfj: If we really want to test the clients... we should make this use querystring for the two lists of scopes we had before...
Just say the querystring values have to JSON encoded lists of scopes. That should be easy to play with...

I suspect most of our client libraries either don't support querystring options yet, or don't have serious tests for them... And certainly not in combination with bewits...
We could say that if no query string is provided, then these hardcoded defaults are assumed.

That way the end-point can be used for testing both cases...


----
Okay, I guess in that way, we can just extend this end-point later... No need to do it now...

Note: query string parameters are always optional..

jonasfj:
`r+`

djmitche:
I think this is pretty ugly, embedding a (single) scope as parameter, but it works and avoids issues with clients that don't yet support queries.

djmitche:
Ideally, we'd have `?clientScope=..&clientScope=..` etc. Same for `requiredScope`.

One option is to just hard-code the clientScope and requiredScope here, and later allow those to be overridden by query args (since not all clients support query args yet).  That would allow easy expansion later without breaking existing users of the endpoint.

jonasfj:
`r+`

PR #48, opened 2016-03-23
Bug 1256743: Remove 'assume:client-id:..' special-casing

djmitche:
The code changes are pretty simple.  The tests turned out to assume this special-case quite prolifically.  I replaced it with thing-id in many cases, just to avoid confusion.

It's possible that this made some of the tests redundant -- I didn't look too hard.  But they all pass for reasons that make sense.


jonasfj:
`r+`

PR #49, opened 2016-03-24
Use new schema validation library

imbstack:
The major change here is that the new validation library is a bit more strict than the old one. I'll call out any particular changes I noticed in comments on the actual line of the changes, but we should decide whether or not this is an ok change to make before this is shipped.


schemas/authenticate-hawk-request.yml
```diff
@@ -43,7 +43,9 @@ properties:
       **Note,** order of querystring elements is important.
   host:
     type:                 string
-    format:               hostname
+    oneOf:
```
imbstack: ajv is more strict about what it considers a hostname

```js
> var ajv = /^[a-z](?:(?:[-0-9a-z]{0,61})?[0-9a-z])?(\.[a-z](?:(?:[-0-9a-z]{0,61})?[0-9a-z])?)*$/i;
undefined
> var jjv = /^([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])(\.([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]{0,61}[a-zA-Z0-9]))*$/;
undefined
> ajv.test('localhost')
true
> jjv.test('localhost')
true
> ajv.test('127.0.0.1')
false
> jjv.test('127.0.0.1')
true
```

test/schemas/create-client-request-no-scopes.json
```diff
@@ -1,4 +1,4 @@
 {
-  "expires": "2000-02-01 00:00:00",
+  "expires": "2016-03-24T00:48:37.942Z",
```
imbstack: ajv is also more strict about what it considers to be a valid `date-time`.

bin/server.js
```diff
@@ -12,6 +12,7 @@ var ScopeResolver      = require('../auth/scoperesolver');
 var signaturevalidator = require('../auth/signaturevalidator');
 var taskcluster        = require('taskcluster-client');
 var url                = require('url');
+var validator_         = require('taskcluster-lib-validate');
```
djmitche: could this be `validate` or if that's taken, `lib_validate`, instead?

djmitche:
This looks good, and as we discussed I'm not too worried about the stricter validation in this particular service.  You can change the variable name or not as you see fit :)

auth/exchanges.js
```diff
@@ -1,8 +1,8 @@
-var base      = require('taskcluster-base');
+var Exchanges = require('pulse-publisher');
```
jonasfj: From what I can see we did bump the version...

package.json
```diff
@@ -27,20 +27,24 @@
     "kexec": "^1.2.0",
     "lodash": "^3.9.3",
     "promise": "^7.0.4",
+    "pulse-publisher": "^0.9.1",
```
jonasfj: Shouldn't we bump tc-base instead?

PR #50, opened 2016-03-25
Bug 1254733: include cert scopes in error message

djmitche:
When handling temporary credentials where cert.scopes are not satisfied
by the issuer's scopes, include the cert scopes in the error message to
help debugging.

### DIFF
--- a/auth/signaturevalidator.js
+++ b/auth/signaturevalidator.js
@@ -101,7 +101,9 @@ var limitClientWithExt = function(credentialName, issuingClientId, accessToken, 
     // Validate certificate scopes are subset of client
     if (!utils.scopeMatch(scopes, [cert.scopes])) {
       throw new Error("ext.certificate issuer `" + issuingClientId +
-                      "` doesn't have sufficient scopes");
+                      "` doesn't satisfiy all certificate scopes " +
+                      cert.scopes.join(', ') + ".  The temporary " +
+                      "credentials were not generated correctly.");
     }
 
     // Generate certificate signature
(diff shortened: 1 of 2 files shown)


auth/signaturevalidator.js
```diff
@@ -101,7 +101,9 @@ var limitClientWithExt = function(credentialName, issuingClientId, accessToken,
     // Validate certificate scopes are subset of client
     if (!utils.scopeMatch(scopes, [cert.scopes])) {
       throw new Error("ext.certificate issuer `" + issuingClientId +
-                      "` doesn't have sufficient scopes");
+                      "` doesn't satisfiy all certificate scopes " +
+                      cert.scopes.join(', ') + ".  The temporary " +
+                      "credentials were not generated correctly.");
```
jonasfj: This is awesome...

One day we probably should use some template language to these errors, so that they can be multi-line, markdown, have be very explanatory.. Like the ones we recently added to tc-lib-api... And I mean for all the errors we report here, better messages would be good...

But this goes very very far to get things right...

jonasfj:
I think pmoore is PTO early next week, so r+

petemoore:
lgtm, thanks for picking this up @djmitche!

auth/signaturevalidator.js
djmitche: If you look at the error in the bug, we're doing a lot of encoding of the json payload in the `message` attribute of another payload.  Not for fixing here, but .. yeah, we could use some cleanup.

PR #51, opened 2016-03-28
Added sentry manager

jonasfj:
Sentry management...

Now we can grant access to sentry using taskcluster credentials.

Also we get: rotating temporary sentry credentials..  

Most notably, this makes it a lot easier to give an sentry access to all workers. As we won't have to burn the secret into the worker AMI... Or manually create a project per workerType and copy/paste credentials around.. 

@djmitche, cc'ing you in as you have bit of a stake in auth too... And this will add another proctype, that should run as scheduled job once a day or so...

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,2 @@
-web: babel-node bin/server.js
+web:            babel-node bin/server.js server
+expireSentry:   babel-node bin/server.js expire-sentry
--- a/auth/v1.js
+++ b/auth/v1.js
@@ -57,10 +57,12 @@ var api = new API({
     table:      /^[A-Za-z][A-Za-z0-9]{2,62}$/,
 
     // Patterns for AWS
-    bucket:     /^[a-z0-9][a-z0-9-]{1,62}[a-z0-9]$/
+    bucket:     /^[a-z0-9][a-z0-9-]{1,62}[a-z0-9]$/,
                 // we could allow "." too, but S3 buckets with dot in the name
                 // doesn't work well with HTTPS and virtual-style hosting.
                 // Hence, we shouldn't encourage people to use them
+    // Project for sentry (and other per project resources)
+    project:    /^[a-zA-Z0-9_-]{1,22}$/,
   },
   context: [
     // Instances of data.Client and data.Role
@@ -79,7 +81,10 @@ var api = new API({
     'azureAccounts',
 
     // Signature validator
-    'signatureValidator'
+    'signatureValidator',
+
+    // SentryManager from sentrymanager.js
+    'sentryManager',
   ]
 });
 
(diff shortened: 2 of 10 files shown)


Procfile
```diff
@@ -1 +1,2 @@
-web: babel-node bin/server.js
+web:            babel-node bin/server.js server
+expireSentry:   babel-node bin/server.js expire-sentry
```
djmitche: Does this end up being useful?  As far as I can tell, the scheduler requires an explicit command, rather than something in Procfile

auth/v1.js
```diff
@@ -57,10 +57,12 @@ var api = new API({
     table:      /^[A-Za-z][A-Za-z0-9]{2,62}$/,
 
     // Patterns for AWS
-    bucket:     /^[a-z0-9][a-z0-9-]{1,62}[a-z0-9]$/
+    bucket:     /^[a-z0-9][a-z0-9-]{1,62}[a-z0-9]$/,
                 // we could allow "." too, but S3 buckets with dot in the name
                 // doesn't work well with HTTPS and virtual-style hosting.
                 // Hence, we shouldn't encourage people to use them
+    // Project for sentry (and other per project resources)
+    project:    /^[a-zA-Z0-9_-]{1,22}$/,
```
djmitche: Is this the same "project" that I've been yammering on about?

djmitche:
This is cool :)

Procfile
jonasfj: I hope so... Because it's what I do in the queue...

I haven't really looked at the logs so I'm not 100% sure it works :)
well, it been a while since I've seen the logs...they are hard to find because they are prefixed heroku scheduler something...

jonasfj: Note: pretty sure heorku docs says you can ref the procfile...

jonasfj: But we do need to add metrics so we can alert if these scheduled jobs don't run

auth/v1.js
jonasfj: Nope, and maybe I should call it sentryProject instead
..
I think I'll use the same pattern for statsum projects though...

Note: it's evident when this shows up in API methods...

Note: if an scm project wanted sentry access, we could grant them a scope... But we would have to manually add their users to sentry, as admins or create a team for them and transfer sentry projects to that...

So this won't make sentry 100% self-service, but will make authentication easier to automate..

djmitche: let's use a different term, then, or at least qualify it

imbstack:
This looks good to me. r+

auth/v1.js
jonasfj: This is just the name of the pattern when it shows up in a url pattern. Nothing more...
It carries no semantics... And outside of the API end-points where it is used it's not visible.

I would like to keep it as project, because I'm adding another end-point that will be using the same pattern, and the same name "project", for a url-pattern parameter.

PR #53, opened 2016-03-28
Added method to grant a statsum token

jonasfj:
Ready to be used with: https://github.com/jonasfj/node-statsum


schemas/statsum-token-response.yml
```diff
@@ -0,0 +1,34 @@
+$schema:  http://json-schema.org/draft-04/schema#
+title:                      "Statsum Token Response"
+description: |
+  Token for submitting statistics to statsum.
+type:                       object
+properties:
```
imbstack: This is spelled correctly this time!

imbstack:
r+ whenever you're ready to ship

schemas/statsum-token-response.yml
jonasfj: I actually had to correct it... after I saw the first bug..

PR #55, opened 2016-03-30
Make smaller scope for sentry

imbstack:
This is a shorter and more straightforward scope.


jonasfj:
@imbstack, travis fails... not sure if it's related... otherwise r+...

jonasfj:
probably more related to tc-client changes...

PR #56, opened 2016-04-06
Fixes for generic sentry issues.

imbstack:
There is a csp field in the dsn returned from the sentry api that we have to get rid of so that it matches the schema. I guess we could include it in the schema instead, but this seems like the more appropriate option.


jonasfj:
r+

PR #58, opened 2016-04-07
node 5 & npm shrinkwrap & compiled babel

imbstack:
This is almost assuredly going to break when we deploy it in some way or another, but it does get us shrinkrwapped packages. Things I've done to see if this works so far:

* the tests pass
* `npm run checkStaging` and `npm run checkStagingSetup` still work
* I've remembered to update the Procfile this time

Things I'm not really sure about:

* I've left a couple of things that really did seem like scripts in `bin/`. Would it make sense for them to be moved to `src/` as well? (`load-test.js` and `make-check-client.js`)

Roping @jonasfj in as well in case he has any useful input here.


djmitche:
Looks good to me, but I feel like what I did looked good too and didn't work, so I'll defer to Jonas

jonasfj:
# r+

Looks really good...

PR #59, opened 2016-04-18
Use upstream rather than jonas' fork

jonasfj:
My branch was merged upstream :)

Note: this might need testing... So perhaps this PR is just as much a TODO item suggesting that we have to do this...


imbstack:
r+

we can work together one day to roll it out and see if it is still functioning.

jonasfj:
Looks like this got fix...

PR #60, opened 2016-04-21
Improve readability

jonasfj:
Just because this is sort of important to get right...

And it was hard to read before... Please triple check that I didn't just introduce a bug.


imbstack:
:ship: r+

imbstack:
This way is approximately 500x more readable :P

PR #62, opened 2016-05-04
STS credentials allow PutObjectAcl, as well

djmitche:
This also required updating the `taskcluster-auth-s3-buckets` policy to
allow PutObjectAcl.

### DIFF
--- a/src/aws.js
+++ b/src/aws.js
@@ -18,6 +18,11 @@ api.declare({
     "parameter is required in the scope guarding access.  The bucket name must",
     "not contain `.`, as recommended by Amazon.",
     "",
+    "Read-only access allows GetObject and ListBucket, limited by the prefix,",
+    "and GetBucketLocation on the bucket.  Read-write access adds PutObject",
+    "and DeleteObject, as well as PutObjectAcl with X-AMZ-ACL set to either",
+    "`public-read` or `private`.",
+    "",
     "The credentials are set to expire after an hour, but this behavior is",
     "subject to change. Hence, you should always read the `expires` property",
     "from the response, if you intend to maintain active credentials in your",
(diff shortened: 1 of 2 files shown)


djmitche:
Updated to limit the allowed x-amz-acl's for PutObjectAcl to just private or public-read (plus docs)

src/aws.js
```diff
@@ -71,55 +76,92 @@ api.declare({
...
+      ],
+      Resource: [
+        'arn:aws:s3:::{{bucket}}/{{prefix}}*'
+          .replace('{{bucket}}', bucket)
+          .replace('{{prefix}}', prefix)
+      ],
+      Condition: {
+        StringEquals: {
+          's3:x-amz-acl': ['private', 'public-read']
+        }
```
jonasfj: I think we should make this restriction on "s3:PutObject" too...
Maybe it should be `['', 'private', 'public-read']` (I'm just guessing that empty string is equal to header missing).

I suspect we'll also want to forbid:
```
x-amz-grant-read	
x-amz-grant-write	
x-amz-grant-read-acp	
x-amz-grant-write-acp	
x-amz-grant-full-control	
```
ie. require the headers are empty...

But I'm not even sure that's enough... because it is possible to specify the ACL in the payload instead of the header...

jonasfj:
@djmitche, the more I look at this it seems like a deep rabbit hole... Where we won't really be sure if people can extend write access...
It would sort of be bad if someone extended write access to say sscache, or granted write access to another S3 account and they lay ready to poison at the opportune time.

Mostly because we wouldn't really notice such a subtle change...

Hence, I'm inclined to ask is this really a good idea? Do we really need to set ACLs...

djmitche:
I don't see a way to forbid an in-body policy being set with PutObjectAcl.  Amazon's policy language is really clunky :(

gregarndt:
For some reason I thought this was fixed, but now I see we have determined that it's not worth the effort.  The only place I know of that we currently would like to set ACLs is with sccache which sets the objects to public-read.  Not being able to set this means that taskcluster will need to use a different tooltool manifest specifying a different sccache package that does not set this.  Not a huge deal, just adds some more configs and mozharness overrides. (unless there is a better way to do this)

djmitche:
I'm still waiting for an answer from Amazon on this.  But adding a feature to sccache to enable/disable this might be a good idea.

jonasfj:
The reason we didn't allow PutObjectAcl
Is that it can be used to allow anyone to overwrite the object... or make
an object public.

Basically, it allows for authority persistence. Well, maybe, let's frame it
as that is what I fear :)
Den 15. jun. 2016 10.07 AM skrev "Dustin J. Mitchell" <
notifications@github.com>:

djmitche:
Right, and the open question with AWS is how to prevent that.

Dustin

PR #63, opened 2016-05-12
Bug 1264100: include information on delegating bucket access

djmitche:


src/aws.js
```diff
@@ -18,6 +18,36 @@ api.declare({
     "parameter is required in the scope guarding access.  The bucket name must",
     "not contain `.`, as recommended by Amazon.",
     "",
+    "This method can only allow access to a whitelisted set of buckets.  To add",
+    "a bucket to that whitelist, contact the TaskCluster team, who will add it to",
+    "the appropriate IAM policy.  If the bucket is in a different AWS account, you",
+    "will also need to add a bucket policy allowing access from the TaskCluster",
+    "account.  That policy should look like this:",
+    "",
+    '    {',
```
jonasfj: I would wrap with:
```js
[
  "```js",
  "  {",
  .... // Example code
  "```",
]
```

Then you are sure to get the code environment.. And you'll get js highlighting which works great for json :)

jonasfj:
# `r+`

PR #64, opened 2016-07-06
Added tc-lib-monitor support

reznord:
Added tc-lib-monitor to package.json & shirnkwrapped

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "url": "https://github.com/taskcluster/taskcluster-auth.git"
   },
   "dependencies": {
+    "assume": "^1.3.1",
     "aws-sdk": "^2.2.0",
     "aws-sdk-promise": "^0.0.2",
     "azure-entities": "^0.9.0",
@@ -33,6 +34,7 @@
     "hawk": "2.3.0",
     "hoek": "^2.16.3",
     "lodash": "^3.9.3",
+    "mocha": "^2.2.5",
     "promise": "^7.0.4",
     "pulse-publisher": "^0.9.1",
     "raven": "^0.10.0",
@@ -41,18 +43,17 @@
     "statsum": "^0.3.0",
     "superagent": "^1.1.0",
     "superagent-hawk": "0.0.4",
+    "superagent-promise": "^0.2.0",
     "taskcluster-base": "^0.12.2",
     "taskcluster-client": "^0.23.9",
     "taskcluster-lib-api": "^0.11.0",
     "taskcluster-lib-app": "^0.8.10",
     "taskcluster-lib-loader": "^0.1.1",
+    "taskcluster-lib-monitor": "^2.2.1",
     "taskcluster-lib-scopes": "^0.8.8",
+    "taskcluster-lib-testing": "^0.11.1",
     "taskcluster-lib-validate": "^0.4.3",
-    "tc-rules": "^5.0.0",
-    "mocha": "^2.2.5",
-    "assume": "^1.3.1",
-    "superagent-promise": "^0.2.0",
-    "taskcluster-lib-testing": "^0.11.1"
+    "tc-rules": "^5.0.0"
   },
   "engines": {
     "node": "^5.8.0",
--- a/src/main.js
+++ b/src/main.js
@@ -11,7 +11,7 @@ let signaturevalidator = require('./signaturevalidator');
 let taskcluster        = require('taskcluster-client');
 let url                = require('url');
 let validate           = require('taskcluster-lib-validate');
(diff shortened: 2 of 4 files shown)


imbstack:
That's a good start! Now it needs to be added to the loader in `main.js`. It might get a bit tricky doing this in auth because normally `tc-lib-monitor` talks to auth to get the keys it needs. I suppose it can talk to itself here? I'm not really sure. This might be getting complicated quickly. Regardless, add it to `src/main.js` and we can go from there.

src/main.js
```diff
@@ -35,23 +34,15 @@ let load = loader({
...
     requires: ['cfg', 'drain'],
-    setup: ({cfg, drain}) => base.stats.startProcessUsageReporting({
-      drain:      drain,
-      component:  cfg.app.statsComponent,
-      process:    'server'
+    setup: ({cfg, drain}) => base.monitor({
+      project:      'taskcluster-auth',
+      drain:        drain,
+      credentials:  cfg.taskcluster.credentials,
+      process:      'server',
```
imbstack: I don't think this should be set to `'server'` manually even if it was before. The `loader` can take `process` in `requires` and in `setup` the same way as `cfg` and `drain`. Then you can just pass it in here!

src/main.js
```diff
@@ -35,23 +34,15 @@ let load = loader({
...
   monitor: {
     requires: ['cfg', 'drain'],
-    setup: ({cfg, drain}) => base.stats.startProcessUsageReporting({
-      drain:      drain,
-      component:  cfg.app.statsComponent,
-      process:    'server'
+    setup: ({cfg, drain}) => base.monitor({
+      project:      'taskcluster-auth',
+      drain:        drain,
+      credentials:  cfg.taskcluster.credentials,
```
imbstack: Here is the problem I was mentioning earlier. I'm fairly sure this won't be available in the config because this is auth itself. This may require us to modify `tc-lib-monitor` somewhat. I'm not sure yet.

imbstack:
Nice progress. This will also need to be passed into all of the places in lib-loader where influx and raven used to be.

package.json
```diff
@@ -41,18 +43,17 @@
     "statsum": "^0.3.0",
```
imbstack: We should figure out why statsum is being included in here. With tc-lib-monitor being used I'm guessing it's no longer needed but I could be wrong.

Anup-Allamsetty: yeah!! will look into it now. will refer tc-queue and see, if that is being used or not.

src/main.js
```diff
@@ -5,14 +5,13 @@ let path               = require('path');
...
 let AWS                = require('aws-sdk-promise');
-let raven              = require('raven');
 let exchanges          = require('./exchanges');
 let ScopeResolver      = require('./scoperesolver');
 let signaturevalidator = require('./signaturevalidator');
 let taskcluster        = require('taskcluster-client');
 let url                = require('url');
 let validate           = require('taskcluster-lib-validate');
-let loader             = require('taskcluster-lib-loader');
+let loader             = require('taskcluster-lib-loader')('process');
```
imbstack: There should not be `('process')` at the end here. Does it even run without stack tracing?

Anup-Allamsetty: sorry, forgot to remove that, will update that in the next push.

src/main.js
```diff
@@ -160,6 +150,7 @@ let load = loader({
           signatureValidator,
           sentryManager,
           statsum:            cfg.app.statsum,
+          monitor:            monitor.prefix('api-context'),
```
imbstack: Is this 'api-context' pattern repeated in some other project? I don't think this makes a lot of sense. Is it generally added to the context elsewhere at all?

src/main.js
```diff
@@ -34,24 +33,15 @@ let load = loader({
...
   monitor: {
-    requires: ['cfg', 'drain'],
-    setup: ({cfg, drain}) => base.stats.startProcessUsageReporting({
-      drain:      drain,
-      component:  cfg.app.statsComponent,
-      process:    'server'
+    requires: ['cfg', 'drain', 'process'],
+    setup: ({cfg, process}) => base.monitor({
+      project:      'taskcluster-auth',
+      credentials:  cfg.taskcluster.credentials,
```
imbstack: This is still going to be our biggest issue because these don't exist. We need to figure out a way to get tc-lib-validate to work without these.

src/main.js
Anup-Allamsetty: yes, it is repeated in tc-queue as well, I guess, it is generally added to the context. 
Can you tell me what this api-context does in general?

PR #65, opened 2016-08-03
Manage scopes [proposal]

jonasfj:
Tests still needed...

After handing out scopes today... I'm realizing that we need to solve the scope delegation issue...
I really want everybody in automation to be able to handout try-equivalent scopes... etc...

In particular, I want to give aki authority to manage the scopes his signing-worker requires from tasks... That way we don't have to delegate them... But importantly we could eventually reduce the amount of scopes we are able to grant...

IMO scopes should be gathered from different people who are responsible for the scopes... You like pokemon you've got to catch them all... :)


jonasfj:
I properly also want to allow `auth:manage-scope:<scope>` to add/remove `<scope>` from clients... And possibly create roles and modify descriptions, but I'm less sure about that!

Note: I imagine we develop a client side html/javascript tool, that looks at which `auth:manage-scope:...` scopes you have and based on it gives you a list of roles/clients  and scope associations that you are responsible for.

The idea being that if you have a `auth:manage-scope:..` scope then you've accepted the responsibility of handing out the scope... And the responsibility of periodically review assignments of the scope and possibly revoke things that don't make sense anymore...

src/v1.js
```diff
@@ -628,26 +628,49 @@ api.declare({
...
   let role = await this.Role.load({roleId}, true);
   if (!role) {
     return res.status(404).json({message: "role not found!"});
   }
 
-  // Check that requester has all the scopes added
-  let added = _.without.apply(_, [input.scopes].concat(role.scopes));
-  if (!req.satisfies([added])) {
-    return;
+  // Find scopes added and removed
```
djmitche: All of this arithmetic will need to be done within the modify, to avoid race conditions.

djmitche:
Can you write up a how-to guide for the various ways we might use this?  The patch itself seems reasonable, but I'm not sure exactly which problems it solves.

My concerns are:

 * "then you've accepted the responsibility of handing out the scope... And the responsibility of periodically review assignments of the scope and possibly revoke things that don't make sense anymore..." scares me -- do our users know about those responsibilities?  Do they have the tools to carry them out?  Do they know how?
 * As you mentioned, this does not address clients.  This might be good -- ideally all management is handled via roles anyway, but we probably need some guidelines to make sure people understand that and design things correctly.
 * This adds complexity to the auth service.  So far, it's pretty simple, so this isn't a huge concern.

src/v1.js
jonasfj: Good catch!!! :)

PR #66, opened 2016-08-11
Bug 1274321: return expires from authenticateHawk; r?jonasfj

djmitche:
This is careful to always return the earliest of client.expires and cert.expiry.

### DIFF
--- a/src/v1.js
+++ b/src/v1.js
@@ -759,7 +759,12 @@ api.declare({
     "the secret credentials leave this service."
   ].join('\n')
 }, function(req, res) {
-  return this.signatureValidator(req.body).then(result => res.reply(result));
+  return this.signatureValidator(req.body).then(result => {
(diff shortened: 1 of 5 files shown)


src/v1.js
```diff
@@ -759,7 +759,12 @@ api.declare({
     "the secret credentials leave this service."
   ].join('\n')
 }, function(req, res) {
-  return this.signatureValidator(req.body).then(result => res.reply(result));
+  return this.signatureValidator(req.body).then(result => {
+    if (result.expires) {
```
jonasfj: the property is required... at least according to the schema... but from what I can see it's always there.

jonasfj:
# `r+`

src/v1.js
djmitche: It's only required for `status: "auth-success"`

PR #67, opened 2016-08-22
Update libs

jonasfj:
Did a round of updating a lot of libraries last week... This is just more of that...


jonasfj:
oh, I also removed deprecated stats setup using influxdb... and have PR open for lib-monitor, so we can bootstrap it later: https://github.com/taskcluster/taskcluster-lib-monitor/pull/43

src/main.js
```diff
@@ -5,53 +5,20 @@ let path               = require('path');
...
 let taskcluster        = require('taskcluster-client');
 let url                = require('url');
-let validate           = require('taskcluster-lib-validate');
-let loader             = require('taskcluster-lib-loader');
-let app                = require('taskcluster-lib-app');
 let SentryManager      = require('./sentrymanager');
 
 // Create component loader
-let load = loader({
+let load = base.loader({
```
djmitche: Why go back to using tc-base?  The move to use the individual libs was deliberate :)

djmitche:
I'm pretty firmly in the "let's kill tc-base" camp, so this feels like a regression to me.  Using tc-base means that you must upgrade in exactly the order that the tc-base maintainer did (so, if you want to use tc-lib-foo 2.0, you might also have to use tc-lib-bar 3.0 since they were modified in that order in tc-base).  It also comes with no changelog, so you have no real idea what libraries you might be upgrading when you upgrade tc-base.  I think the most convincing argument I've heard from tc-base is that `var base = require('taskcluster-base')` is a lot more natural than `var lib_api = require('taskclutser-lib-api')` -- and it's not a terribly strong argument :)

src/main.js
jonasfj: oh,

jonasfj:
he he, tc-base is perhaps a bit of a weird construction...

> I think the most convincing argument I've heard from tc-base is that `var base = require('taskcluster-base')` is a lot more natural  -- and it's not a terribly strong argument :)

Okay, I'll admit that's a bad argument :) hehe
So far my argument have been that it was an easy to way to poll everything in... But yes, it's not the greatest argument.

I guess another argument for it is that it forces everybody to upgrade and use the same library versions. Even if it's painful. I don't know if this is best argument either, as I'm effectively arguing for the painful thing.

But by using tc-base, you can still pull in older versions of libraries separately if you wish to. But then you're being very explicit about the fact that you're intentionally not up-to-date.

I guess part of me hopes that one day, tc-base will move more in the direction of being a framework.


----

If we want to kill `taskcluster-base` we definitely need some sane meta-variable names for other modules, and we need to agree in what these should be. I guess `let Name = require('taskcluster-lib-name');` would work, but some of our code would definitely need modifications for this to work.

Perhaps we should reach consensus on whether or not to kill tc-base.@bstack thoughts?

djmitche:
itym @imbstack :)

"it forces everybody to upgrade and use the same library versions" -- it doesn't force everybody to upgrade, which is I think the important thing here.

Using other versions of libraries (not just older -- more often *newer*) is possible, and it's what I've done in several places including here in auth.  So it seems silly to revert those, possibly downgrading them (unless you've very recently updated base), to use the versions in base.

jonasfj:
I guess I sort of also fear that it'll be hard to find libraries... most of the taskcluster-lib-... is easy to discover but typed-env-config, and pulse-publisher is less obvious...

But yes, all of that is probably better solved with some sort of example repository outline how to build services...

djmitche:
https://docs.taskcluster.net/reference/libraries

jonasfj:
good point, I'm sold...

djmitche:
Looks good to me!

PR #70, opened 2016-09-27
DNS -> DSN typo, credits @imbstack

jonasfj:


src/main.js
```diff
@@ -45,7 +45,7 @@ let load = Loader({
             expires: key.expires.toJSON(),
           };
         },
-        sentryDNS: async (project) => {
+        sentryDSN: async (project) => {
           return {
             project,
             token:    Statsum.createToken(project, cfg.app.statsum.secret, '25h'),
```
imbstack: Wait, aren't we passing in Statsum.createToken for sentryDSN?

Did we do this backwards?

jonasfj: yes, we did it backwards :)

PR #76, opened 2016-10-04
Added executeDFA hopefully solving memory issues

jonasfj:
https://gist.github.com/jonasfj/ed38276ec2bde6c66e15c71ffa109d6d

### DIFF
--- a/src/dfa.js
+++ b/src/dfa.js
@@ -410,35 +410,24 @@ let generateDFA = (roles, i, n, k, sets, implied) => {
 exports.generateDFA = generateDFA;
 
 /**
- * Builds a pair {resolver, sets} where sets is a list of lists of roles,
- * and given a scope `resolver(scope)`` returns an index from sets.
- *
- * That definition sounds slightly complicated, it's actually very simple,
- * sets is on the form:
- * ```js
- * sets = [
- *   [{role}, ...],
- *   [{role}, ...],
- *   [{role}, ...]
- * ]
- * ```
- *
- * For efficiency we allow sets[i] = [{role}, ..., j] where j < i to be
- * interpreted as sets[i].concat(sets[j]). By not duplicating we don't have to
- * resolve `sets[j]` multiple times, even though the set appears in multiple
- * other sets.
- *
- * The `resolver` function returns an index in the sets array, as this allows
- * us to later create a new sets variable where roles have been expanded to
- * the scopes they imply, and just like that we can use the `resolver` to go
- * from scope to expanded scopes.
+ * Finds the set identifier given a scope and initial state.
  */
-let buildResolver = (roles) => {
-  // Generate DFA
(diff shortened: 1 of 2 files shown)


djmitche:
Just to summarize what I'm seeing here:

 * generateDFA creates a data structure representing the DFA (no change)
 * executeDFA executes that DFA symbolically
 * compileDFA converts that DFA to a compiled JS function
 * buildResolver now calls executeDFA because it's a little faster, and compileDFA is never called

So I'm guessing that the memory issue has something to do with dynamically compiling functions?  Still, isn't buildResolver only called once per dyno startup, in which case only one function is being dynamically compiled -- how did that eat 1GB+?

If compileDFA isn't used, then it should be removed from the file.

src/dfa.js
```diff
@@ -410,35 +410,24 @@ let generateDFA = (roles, i, n, k, sets, implied) => {
...
+ * Finds the set identifier given a scope and initial state.
  */
-let buildResolver = (roles) => {
-  // Generate DFA
-  roles = sortRolesForDFAGeneration(roles);
-  let sets = [[]];
-  let dfa = generateDFA(roles, 0, roles.length, 0, sets, 0);
+let executeDFA = (state, scope, depth = 0) => {
+  let n = scope.length;
+  if (typeof(state.end) === 'number' && n == depth) {
```
imbstack: Maybe add a comment here about why we have to check if it's a number. It's because of the case where the initial character isn't a match at all, right? It is kinda non-obvious.

jonasfj:
> So I'm guessing that the memory issue has something to do with dynamically compiling functions? Still, isn't buildResolver only called once per dyno startup, in which case only one function is being dynamically compiled

Well, we rebuild the resolver every time there is a change (as we get notified by pulse messages). We also rebuild every 20min or so, just in case we missed a pulse message. But that should still be perfectly fine, ie. the GC should cleanup after us.

> how did that eat 1GB+?

It didn't, well, my stats say that the maximum heap used is 1GB, but the real problem is RSS being 9GB... RSS is what heroku will read. I'm sure this has to do with how node GC workings, it probably can't move functions, maybe when they get optimized some of the optimized function code can't be moved... So the GC end up with extreme heap fragmentation, and bad things happen.

When max used is 1GB and RSS is 9GB, there must be some extremely bad heap fragmentation, that's all I can imagine.

> If compileDFA isn't used, then it should be removed from the file.

Hmm... I would like to keep it around as we can enable it in some future if it becomes relevant again.
Or have it as inspiration...

djmitche:
I know it's pretty, but I really dislike having unused code floating around, especially in an important app like this one.  If you really want to see it in the `master` branch, moving it to `misc/compiled-dfa.js` or something might be OK

jonasfj:
okay, move it to a separate file...

PR #77, opened 2016-10-13
Changed ping() description

owlishDeveloper:
This is my fix for bug https://bugzilla.mozilla.org/show_bug.cgi?id=1262612 .

### DIFF
--- a/src/v1.js
+++ b/src/v1.js
@@ -877,11 +877,11 @@ api.declare({
   method:   'get',
   route:    '/ping',
   name:     'ping',
-  stability:  'experimental',
-  title:    "Ping Server",
+  stability:  'stable',
+  title:    'Ping Server',
   description: [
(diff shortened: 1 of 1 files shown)


owlishDeveloper:
@djmitche r?

src/v1.js
```diff
@@ -880,9 +880,8 @@ api.declare({
   stability:  'experimental',
```
djmitche: Let's change this to 'stable' too

jonasfj:
so wouldn't it be much better if we implemented the ping method in:
https://github.com/taskcluster/taskcluster-lib-api

When we would have the ping method in all systems... and we have the same implemetation everywhere...

owlishDeveloper:
@djmitche r?

djmitche:
deploy'd to production

PR #82, opened 2016-10-21
New ping API, removed old one

owlishDeveloper:
@imbstack @djmitche r?


djmitche:
it's live!

djmitche:
And the ping endpoint is correctly in the reference: http://references.taskcluster.net/auth/v1/api.json

PR #84, opened 2016-11-11
Bug 1316698 - EC2 metadata compat

jonasfj:
https://bugzilla.mozilla.org/show_bug.cgi?id=1316698

### DIFF
--- a/src/aws.js
+++ b/src/aws.js
@@ -6,6 +6,9 @@ api.declare({
   name:       'awsS3Credentials',
   input:      undefined,
   output:     'aws-s3-credentials-response.json#',
+  query: {
+    format:   /iam-role-compat/,
+  },
   deferAuth:  true,
   stability:  'stable',
   scopes:     [['auth:aws-s3:<level>:<bucket>/<prefix>']],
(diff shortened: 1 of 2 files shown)


src/aws.js
```diff
@@ -163,12 +174,26 @@ api.declare({
     DurationSeconds:    60 * 60   // Expire credentials in an hour
   }).promise();
 
+  // Make result compatibility with how EC2 metadata service let's instances
+  // access IAM roles
+  if (req.query.format === 'iam-role-compat') {
+    return res.status(200).json({
```
djmitche: This avoids the schema validation.. clever :)

test/s3_test.js
```diff
@@ -150,4 +150,28 @@ suite('aws S3 (STS)', () => {
...
+
+    let s3 = new aws.S3({
+      accessKeyId:     result.AccessKeyId,
+      secretAccessKey: result.SecretAccessKey,
+      sessionToken:    result.Token,
+    });
+    debug("### s3.putObject");
+    await s3.putObject({
+      Bucket:   bucket,
+      Key:      'folder1/' + id,
```
djmitche: why create a new object on every test run?

djmitche:
```
     Error: getaddrinfo ENOTFOUND taskclusterdev.table.core.windows.net taskclusterdev.table.core.windows.net:443
```
why does DNS suck?

jonasfj:
> why does DNS suck?

Not a DNS issue `taskclusterdev` was my personal account after you leaked creds I just deleted it... So rotating tc-auth tests to use jungle account.

test/s3_test.js
jonasfj: to avoid consistency issues... this uses the garbage bucket where everything expires after 24 hours.

PR #85, opened 2016-11-16
Add tc-lib-docs

imbstack:
as it says on the tin.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "taskcluster-client": "^1.2.0",
     "taskcluster-lib-api": "^3.0.0",
     "taskcluster-lib-app": "^1.0.0",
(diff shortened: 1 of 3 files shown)


imbstack:
Not sure why tests are failing. Looking into it.

package.json
```diff
@@ -41,7 +41,7 @@
     "taskcluster-client": "^1.2.0",
     "taskcluster-lib-api": "^3.0.0",
     "taskcluster-lib-app": "^1.0.0",
-    "taskcluster-lib-docs": "^2.0.0",
+    "taskcluster-lib-docs": "^3.1.0",
```
djmitche: Interesting that it was already included, but not used?

djmitche:
deployed and promoted

PR #86, opened 2017-01-19
Bump pulse-publisher

jonasfj:
Please feel free to land this... Should fix the pulse issue we had tuesday...

### DIFF
--- a/schemas/authenticate-hawk-request.yml
+++ b/schemas/authenticate-hawk-request.yml
@@ -43,9 +43,9 @@ properties:
       **Note,** order of querystring elements is important.
   host:
     type:                 string
-    oneOf:
-      - format:               hostname
-      - format:               ipv4
+    anyOf:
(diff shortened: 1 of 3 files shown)


djmitche:
I'll let you handle the landing, particularly given the schema validation error in Travis

schemas/authenticate-hawk-request.yml
```diff
@@ -43,9 +43,9 @@ properties:
       **Note,** order of querystring elements is important.
   host:
     type:                 string
```
djmitche: does this need to be removed?

jonasfj: Sorry, I got it wrong... I remove the other `type: string`, I don't know if this works... But why wouldn't it, I suspect it does :)

jonasfj:
I'll land when travis passes...

PR #89, opened 2017-01-25
[DO NOT MERGE] Optional autoexpire

djmitche:
This needs tests and testing, but I'd like a 30% review here if you don't mind.


schemas/create-client-request.yml
```diff
@@ -9,6 +9,12 @@ properties:
       Date and time where the clients access is set to expire
     type:                   string
     format:                 date-time
+  deleteOnExpiration:
+    description: |
+      If true, the service may delete this client after it has expired.  If
+      false (the default), the client will remain after expiration, although
+      it cannot be used for authentication in that state.
+    type:                   boolean
```
jonasfj: Add `default: false`

schemas/create-client-request.yml
```diff
@@ -9,6 +9,12 @@ properties:
       Date and time where the clients access is set to expire
     type:                   string
     format:                 date-time
+  deleteOnExpiration:
+    description: |
+      If true, the service may delete this client after it has expired.  If
+      false (the default), the client will remain after expiration, although
```
jonasfj: This will be rendered as markdown, so you can write:
```
If `false` (the default)...
```

schemas/create-client-response.yml
```diff
@@ -20,6 +20,12 @@ properties:
       Date and time where the clients access is set to expire
     type:                   string
     format:                 date-time
+  deleteOnExpiration:
+    description: |
+      If true, the service may delete this client after it has expired.  If
+      false (the default), the client will remain after expiration, although
+      it cannot be used for authentication in that state.
+    type:                   boolean
```
jonasfj: Add `deleteOnExpiration` as required property. This is a response.

jonasfj: Do not document default value here.

schemas/get-client-response.yml
```diff
@@ -14,6 +14,12 @@ properties:
       Date and time where the clients access is set to expire
     type:                   string
     format:                 date-time
+  deleteOnExpiration:
+    description: |
+      If true, the service may delete this client after it has expired.  If
+      false (the default), the client will remain after expiration, although
```
jonasfj: No need to document default value

schemas/get-client-response.yml
```diff
@@ -14,6 +14,12 @@ properties:
       Date and time where the clients access is set to expire
     type:                   string
     format:                 date-time
+  deleteOnExpiration:
```
jonasfj: Property should be listed as required in the `required` property at bottom of object description...

src/data.js
```diff
@@ -26,10 +26,11 @@ var Client = Entity.configure({
...
      * Details object with properties:
-     * - created          // Time when client was created
-     * - lastModified     // Last time client was modified
-     * - lastDateUsed     // Only updated if more than 6 hours out of date
-     * - lastRotated      // Last time accessToken was reset
+     * - created            // Time when client was created
+     * - lastModified       // Last time client was modified
+     * - lastDateUsed       // Only updated if more than 6 hours out of date
+     * - lastRotated        // Last time accessToken was reset
+     * - deleteOnExpiration // If true, can be deleted after expiration
```
jonasfj: Either do a migrate function to next version... or declare that  `deleteOnExpiration` is optional.

src/data.js
```diff
@@ -51,16 +52,17 @@ Client.prototype.expandedScopes = function(resolver) {
...
-    expandedScopes: this.expandedScopes(resolver),
-    disabled:       !!this.disabled
+    clientId:           this.clientId,
+    description:        this.description,
+    expires:            this.expires.toJSON(),
+    created:            this.details.created,
+    lastModified:       this.details.lastModified,
+    lastDateUsed:       this.details.lastDateUsed,
+    lastRotated:        this.details.lastRotated,
+    deleteOnExpiration: this.details.deleteOnExpiration,
```
jonasfj: Add `|| false `, as property is optional in entity. (or just upgrade the entity to next version).

src/data.js
jonasfj: Just add:
```js
configure({
  version:          2,
  signEntities:     true,
  properties: {
    clientId:       Entity.types.String,
    description:    Entity.types.Text,
    accessToken:    Entity.types.EncryptedText,
    expires:        Entity.types.Date,
    /**
     * Details object with properties:
     * - created            // Time when client was created
     * - lastModified       // Last time client was modified
     * - lastDateUsed       // Only updated if more than 6 hours out of date
     * - lastRotated        // Last time accessToken was reset
     * - deleteOnExpiration // If true, can be deleted after expiration
     */
    details:        Entity.types.JSON,
    scopes:         Entity.types.JSON,  // new in v2
    disabled:       Entity.types.Number // new in v2
  },
  migrate(item) {
    item.details = _.defaults({}, item.details, {deleteOnExpiration: false});
    return item;
  }
});
```
Note: when modifying `item` I think we have to create a new value... I don't remember if azure-entities does a clone, it might :)
(Just the property, I'm pretty sure `item` may be modified)

jonasfj: One could argue that we should make `deleteOnExpiration` a top-level property so we can filter on it later... But we're probably not going to have so many expired clients that quering is going to be a problem...

src/main.js
```diff
@@ -213,15 +213,24 @@ let load = Loader({
     }
   },
 
-  'expire-sentry': {
-    requires: ['cfg', 'sentryManager'],
-    setup: async ({cfg, sentryManager}) => {
+  'expire': {
```
jonasfj: I would have made two processes, but I see how both these do a few things... Perhaps we can run them in parallel...


Note: there is a value in having two processes when writing tests.

src/data.js
djmitche: I thought about the filtering, yes, and reached the same conclusion :)

src/main.js
djmitche: I may factor this a bit for tests, and yes, running in parallel makes sense.  But two Heroku dynos seems like overkill for two fairly short jobs that do similar things (expire stuff).  Is there an advantage?  I'm not sure what the Heroku costs are, either..

jonasfj:
Don't worry about heroku costs for this...
It only runs once every 24 hours right?

Or maybe 4 times a day using the heroku scheduler... This creates one-off
dynos... Afaik it's prorated per second, so who cares...
https://devcenter.heroku.com/articles/usage-and-billing

Even if it were per month it's a tiny price for separating concerns...

Den 25. jan. 2017 12.31 PM skrev "Dustin J. Mitchell" <
notifications@github.com>:

djmitche:
Oh, hah, per second.. that's different!

Dustin

djmitche:
OK, I think that addresses everything and adds tests.  Full review, when you have a moment?

src/data.js
```diff
@@ -86,13 +113,37 @@ Client.ensureRootClient = function(accessToken) {
...
+Client.purgeExpired = async function(now = new Date()) {
+  var count = 0;
+  var expired = await this.scan({
+    expires: Entity.op.lessThan(now),
+  }, {
+    limit: 100,
+    handler: async client => {
+      if (client.details.deleteOnExpiration) {
+        count++;
+        console.log("kill", client.clientId);
```
jonasfj: Maybe write `auto-deleted` instead of `kill`

jonasfj:
@djmitche, please consider PR #90 before landing this... If we're going to upgrade to a version 3 in `data.js` I propose that we also move to use the new `Schema` construct :)

Note: test will fail because of missing npm-shrinkwrap updated...

src/data.js
djmitche: Whoops, leftover debugging..

src/data.js
djmitche: (I opted for the upgrade)

djmitche:
Staging checks passed, so pushing to production.

PR #90, opened 2017-02-01
Use Entity.types.Schema to enforce details

jonasfj:
Note: this also need an updated to `npm-shrinkwrap.json`, tests probably won't pass because if it...
This was just a quick hack to outline the idea...


jonasfj:
Okay, I updated npm-shrinkrwap

PR #91, opened 2017-02-03
Add level to azure sas creds

imbstack:
As we discussed for azure-backups.

### DIFF
--- a/src/azure.js
+++ b/src/azure.js
@@ -3,28 +3,36 @@ var api         = require('./v1');
 
 api.declare({
   method:     'get',
-  route:      '/azure/:account/table/:table/read-write',
+  route:      '/azure/:account/table/:table/:level',
   name:       'azureTableSAS',
   input:      undefined,
   output:     'azure-table-access-response.json#',
   deferAuth:  true,
   stability:  'stable',
-  scopes:     [['auth:azure-table-access:<account>/<table>']],
+  scopes:     [['auth:azure-table-access:<account>/<table>'], ['auth:azure-table-access:<account>/<table>/<level>']],
   title:      "Get Shared-Access-Signature for Azure Table",
   description: [
     "Get a shared access signature (SAS) string for use with a specific Azure",
-    "Table Storage table.  Note, this will create the table, if it doesn't",
-    "already exist."
+    "Table Storage table. By not specifying a level as in azureTableSASLevel,",
+    "you will get read-write permissions. If you get read-write from this, it will create the",
+    "table if it doesn't already exist.",
   ].join('\n')
 }, async function(req, res) {
-  // Get parameters
   var account   = req.params.account;
   var tableName = req.params.table;
+  var level     = req.params.level || 'read-write';
+
+  if (!['read-write', 'read-only'].includes(level)) {
+    return res.status(404).json({
+      message:    "Level '" + level + "' is not valid. Must be one of ['read-write', 'read-only']."
+    });
+  }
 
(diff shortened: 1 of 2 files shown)


src/azure.js
```diff
@@ -3,7 +3,7 @@ var api         = require('./v1');
 
 api.declare({
   method:     'get',
-  route:      '/azure/:account/table/:table/read-write',
+  route:      '/azure/:account/table/:table',
```
jonasfj: So we can change this to: `/azure/:account/table/:table/:level` without breaking the REST API, right?

We break the definition of how `auth.azureTableSAS` works in future versions of the client libraries.
So if we want to do the pretty thing and built least possible legacy, what we do is we break the client libraries, but not the REST API. Hence, older client libraries will still work, they just don't have the parameter.

Given that this is a method used in a very few places I don't think this is a problem. As long as we upgrade the node client to version 2, after we update the reference docs this should be fine. I doubt anyone using the python or golang clients are using `azureTableSAS`, but sure we should bump those too...

Note: I guess in general this might be a little bad practice... but given that this method is used few places, I think we should welcome the chance to avoid legacy in the name of backward compatibility.

imbstack:
Isn't this used anywhere that azure-entities is used without passing in azure creds directly? That seems like a pretty common pattern and one that will be a bit hard to root out all of the occurences of? Otherwise I agree.

src/azure.js
imbstack: Isn't this used anywhere that azure-entities is used without passing in azure creds directly? That seems like a pretty common pattern and one that will be a bit hard to root out all of the occurences of? Otherwise I agree.

imbstack: I guess https://tools.taskcluster.net/auth/scopes/#auth:azure-table-access/ (with the sub-scope option) is the full list of things that have the scope to use this?

src/azure.js
```diff
@@ -3,28 +3,36 @@ var api         = require('./v1');
...
   method:     'get',
-  route:      '/azure/:account/table/:table/read-write',
+  route:      '/azure/:account/table/:table/:level',
   name:       'azureTableSAS',
   input:      undefined,
   output:     'azure-table-access-response.json#',
   deferAuth:  true,
   stability:  'stable',
-  scopes:     [['auth:azure-table-access:<account>/<table>']],
+  scopes:     [['auth:azure-table-access:<account>/<table>'], ['auth:azure-table-access:<account>/<table>/<level>']],
```
jonasfj: let's use `auth:azure-table:<level>:<account>/<table>`

src/azure.js
```diff
@@ -3,28 +3,36 @@ var api         = require('./v1');
...
-    "already exist."
+    "Table Storage table. By not specifying a level as in azureTableSASLevel,",
+    "you will get read-write permissions. If you get read-write from this, it will create the",
+    "table if it doesn't already exist.",
   ].join('\n')
 }, async function(req, res) {
-  // Get parameters
   var account   = req.params.account;
   var tableName = req.params.table;
+  var level     = req.params.level || 'read-write';
```
jonasfj: No need for the `|| 'read-write'`, the old clients have it embedded in their path.

jonasfj:
@imbstack, I think this is ready to ship.

@elenasolomon, this was the scope change I mentioned in the email... I suspect a similar pattern would be great for azure blob sas stuff.. Note: the `auth:azure-table-access:` pattern will probably be removed in the future, in favor of the `auth:azure-table:read-write:..` pattern..

PR #92, opened 2017-02-03
Add table listing abilities to azure manager

imbstack:
Also add some table listing stuff for backups.

### DIFF
--- a/schemas/azure-table-list-response.yml
+++ b/schemas/azure-table-list-response.yml
@@ -1,0 +1,28 @@
+$schema:  http://json-schema.org/draft-04/schema#
+title:                      "Azure List Account Response"
+description: |
+  A list of Azure tables in an account
+type:                       object
+properties:
+  tables:
+    title: Azure Tables
+    description: |
+      A list of tables that are in an account. These are
+      the tables that can have SAS credentials fetched for them.
+    type:                 array
+    items:
+      type: string
+  continuationToken:
+    type:             string
+    title:            Continuation Token
+    description: |
+      Opaque `continuationToken` to be given as query-string option to get the
+      next set of tables.
+      This property is only present if another request is necessary to fetch all
+      results. In practice the next request with a `continuationToken` may not
+      return additional results, but it can. Thus, you can only be sure to have
+      all the results if you've called `azureAccountTables` with `continuationToken`
+      until you get a result without a `continuationToken`.
+additionalProperties:       false
+required:
+  - tables
--- a/src/azure.js
+++ b/src/azure.js
@@ -1,5 +1,56 @@
+var _           = require('lodash');
 var azure       = require('fast-azure-storage');
 var api         = require('./v1');
+
+api.declare({
+  method:     'get',
+  route:      '/azure/accounts',
+  name:       'azureAccounts',
+  input:      undefined,
+  output:     'azure-account-list-response.json#',
+  stability:  'stable',
+  scopes:     [['auth:azure-table:list-accounts']],
+  title:      "List Accounts Managed by Auth",
+  description: [
+    "Retrieve a list of all Azure accounts managed by Taskcluster Auth.",
+  ].join('\n')
+}, function(req, res) {
+  return res.reply({accounts: _.keys(this.azureAccounts)});
+});
+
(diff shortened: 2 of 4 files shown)


schemas/azure-table-list-response.yml
```diff
@@ -0,0 +1,24 @@
...
+properties:
+  tables:
+    title: Azure Tables
+    description: |
+      A list of tables that are in an account. These are
+      the tables that can have SAS credentials fetched for them.
+    type:                 array
+    items:
+      type: string
+  nextTableName:
```
jonasfj: Let call it `continuationToken` and just say it's an opaque token.

src/azure.js
```diff
@@ -1,8 +1,56 @@
...
+}, function(req, res) {
+  return res.reply({accounts: _.keys(this.azureAccounts)});
+});
+
+api.declare({
+  method:     'get',
+  route:      '/azure/:account/tables',
+  name:       'azureAccountTables',
+  query: {
+    nextTableName: /./,
```
jonasfj: Again list call it continuation token... and do a proper regex: `/^[A-Za-z][A-Za-z0-9]{2,62}$/`.
The regex is actually from: https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/understanding-the-table-service-data-model

imbstack:
The only issue with `[{account: '...', table: '...'}, ...]` is that the continuationToken stops making as much sense. I agree that we should change the name. I'll have it up shortly.

imbstack:
@jonasfj: Ok, this is up-to-date with master now. I still think the two functions kinda make sense for the same reasons as my previous comment, but am open to suggestions. Does this look reasonable to you?

src/azure.js
```diff
@@ -1,8 +1,61 @@
+var _           = require('lodash');
 var azure       = require('fast-azure-storage');
 var api         = require('./v1');
 
 api.declare({
   method:     'get',
+  method:     'get',
+  route:      '/azure/:account/table/:table/read-write',
+  route:      '/azure/accounts',
```
jonasfj: two routes?

src/azure.js
```diff
@@ -1,8 +1,61 @@
...
+  name:       'azureAccounts',
+  input:      undefined,
+  output:     'azure-account-list-response.json#',
+  stability:  'stable',
+  scopes:     [['auth:azure-table-access:list-accounts']],
```
jonasfj: I propose scope patterns:
```
auth:azure-table:<level>:<account>/<table>, where level is 'read-only' or 'read-write'
auth:azure-table:list-accounts
auth:azure-table:list-tables:<account>
```

src/azure.js
```diff
@@ -1,8 +1,61 @@
...
+    "Retrieve a list of all Azure accounts managed by Taskcluster Auth.",
+  ].join('\n')
+}, function(req, res) {
+  return res.reply({accounts: _.keys(this.azureAccounts)});
+});
+
+api.declare({
+  method:     'get',
+  route:      '/azure/:account/tables',
+  name:       'azureAccountTables',
```
jonasfj: How about `azureTables`

src/azure.js
```diff
@@ -1,8 +1,61 @@
...
+    accountId:  account,
+    accessKey:  this.azureAccounts[account]
+  });
+
+  let result = await table.queryTables({continuationToken});
+  if (result.nextTableName) {
+    result.continuationToken = result.nextTableName;
+  }
+  delete result.nextTableName;
+  return res.reply(result);
```
jonasfj: prefer:
```js
let data = {tables: result.tables};
if (result.nextTableName) {
  data.continuationToken = result.nextTableName;
}
return res.reply(data);
```
Just as long, but you won't have an error when/if we add properties in `fast-azure-storage`.

IMO, we should augment `res.reply` so that it removes properties that has the value `undefined`. If we serialized to JSON they wouldn't be there either... The only reason we can just do: `res.reply({tables: result.tables, continuationToken: result.nextTableName || undefined});` is because `ajv` complains about `undefined`... Which is stupid because when serialized `{a: undefined}` becomes `"{}"`

PR #94, opened 2017-02-10
Azure blob access

elenasolomon:
Adds an API endpoint for getting a SAS string for use with a specific Azure blob storage container

Please do not merge it, because the implementation of Blob service is not yet merged.


elenasolomon:
@jonasfj based on your last email, I changed the scope to:
auth:azure-blob:<level>:<account>/<container>

src/azure.js
```diff
@@ -69,4 +69,89 @@ api.declare({
...
+      container:    containerName,
+      level:        level,
+    })) {
+    return;
+  }
+
+  // Check that the account exists
+  if (!this.azureAccounts[account]) {
+    return res.status(404).json({
+      message:    "Account '" + account + "' not found, can't delegate access"
```
jonasfj: Use res.reportError I think it's called... See tc-lib-api

src/azure.js
```diff
@@ -69,4 +69,89 @@ api.declare({
...
+    "if it doesn't already exists."
+  ].join('\n')
+}, async function(req, res){
+  // Get parameters
+  let account = req.params.account;
+  let containerName = req.params.container;
+  let level = req.params.level;
+
+  if (['read-write', 'read-only'].indexOf(level) < 0) {
+    return res.status(404).json({
```
jonasfj: Use res.reportError

src/azure.js
```diff
@@ -69,4 +69,89 @@ api.declare({
...
+    start:         new Date(Date.now() - 15 * 60 * 1000),
+    expiry:        expiry,
+    resourceType: 'container',
+    permissions: {
+      read:       true,
+      add:        perm,
+      create:     perm,
+      write:      perm,
+      delete:     perm,
+      list:       perm
```
jonasfj: Is list not read only? I'm not sure what it covers...

elenasolomon: You are right. 'List' gives the permission to list blobs in the container.

jhford:
@elenasolomon can you please rebase your patch when you have a chance.  You'll also need to update the package.json to match the new version of fast-azure-storage.

@jonasfj do the review addressing commits look good?  Once we land fast-azure-storage with blob storage, would you be OK with this being merged?

elenasolomon:
I made the rebase and I changed also the version of 'fast-azure-storage' library

schemas/azure-blob-response.yml
```diff
@@ -0,0 +1,22 @@
+$schema:  http://json-schema.org/draft-04/schema#
+title:    "Azure Shared-Access-Signature Response"
```
jonasfj: Included `Blob` and `Table` in the two titles for this schema and the table schema... Afaik golang client uses the titles to generate types :)

schemas/azure-blob-response.yml
```diff
@@ -0,0 +1,22 @@
...
+    type:                 string
+  expiry:
+    description: |
+      Date and time of when the Shared-Access-Signature expires.
+    type:                   string
+    format:                 date-time
+additionalProperties:       false
+required:
+  - sas
+  - expiry
```
jonasfj: end document with line break

src/azure.js
```diff
@@ -82,3 +82,86 @@ api.declare({
     expiry:   expiry.toJSON()
   });
 });
+
+api.declare({
+  method:     'get',
+  route:      '/azure/:account/blob/:container/:level',
```
jonasfj: How about `'/azure/:account/containers/:container/:level',`

The other one we use: `/azure/<account>/table/<table>/<level>`
And we probably should have used `/azure/<account>/tables/<table>/<level>`. So that matches with the route for listing.

jonasfj: Also we should define a pattern for the `container` parameter here:
https://github.com/elenasolomon/taskcluster-auth/blob/d1eb601a2b9b5f2709f4c0f672a7d060d99169d1/src/v1.js#L57

Perhaps we can even define `level` there too... Since level is always `/^(read-write|read-only)$/`

src/azure.js
```diff
@@ -82,3 +82,86 @@ api.declare({
...
+    return res.reportError('InvalidRequestArguments',
+      `Level '${level}' is not valid. Must be one of ['read-write', 'read-only'].`);
+  }
+
+  // Check that the client is authorized to access given account and container
+  if (!req.satisfies({
+      account:      account,
+      container:    containerName,
+      level:        level,
+    })) {
```
jonasfj: Please use the short hand: `req.satisfies({account, container: containerName, level})`
If you changed `containerName` to `container` it would work even better :)

jonasfj: Actually, a better test is:
`level === 'read-only' && req.satisfies({account, container: containerName, level: 'read-write'}, true) && req.satisfies({account, container: containerName, level})`

That way, the 'read-write' scope grants 'read-only' too...

Credits @imbstack for remembering this for tables...

schemas/azure-table-access-response.yml
```diff
@@ -1,5 +1,5 @@
 $schema:  http://json-schema.org/draft-04/schema#
-title:                      "Azure Shared-Access-Signature Response"
+title:                      "Azure Table Shared-Access-Signature Response"
```
jonasfj: Thanks :)

src/azure.js
```diff
@@ -88,9 +88,8 @@ api.declare({
 
   // Check that the account exists
   if (!this.azureAccounts[account]) {
-    return res.status(404).json({
-      message:    "Account '" + account + "' not found, can't delegate access"
-    });
+    return res.reportError('InvalidRequestArguments',
```
jonasfj: ResourceNotFound  produces 404

src/azure.js
```diff
@@ -131,3 +130,84 @@ api.declare({
...
+  // Check that the client is authorized to access given account and container
+  if (!(level === 'read-only' &&
+    req.satisfies({account, container, level: 'read-write'}, true)) &&
+    !req.satisfies({account, container, level})) {
+    return;
+  }
+
+  // Check that the account exists
+  if (!this.azureAccounts[account]) {
+    return res.reportError('InvalidRequestArguments',
```
jonasfj: Again this is probably a not found error

test/azure_test.js
```diff
@@ -181,4 +180,133 @@ suite('azure table (sas)', function() {
...
+    assert(typeof(result.sas) === 'string', "Expected some form of string");
+    assert(new Date(result.expiry).getTime() > new Date().getTime(),
+      "Expected expiry to be in the future");
+  });
+
+  test('azureBlobSAS (unauthorized container)', async () => {
+    let auth = new helper.Auth({
+      baseUrl:          helper.baseUrl,
+      credentials:      rootCredentials,
+      authorizedScopes: [
```
jonasfj: Helper has a quick method to set authorised scopes...

elenasolomon: Which method? Sorry, but I do not find a method that can set authorizedScopes

jonasfj: oh, we don't have it here...
In queue we do: https://github.com/taskcluster/taskcluster-auth/blob/ce1f509c11179875674a7c548ca07fd0e8db20ac/test/helper.js#L76-L82

We could do the same here: 
https://github.com/taskcluster/taskcluster-auth/blob/ce1f509c11179875674a7c548ca07fd0e8db20ac/test/helper.js#L76-L82

Ie. basically wrap with  a `helper.scopes` method...

jonasfj:
@elenasolomon let us know when this isn't **"[DO-NOT-MERGE]"** anymore...

elenasolomon:
@jonasfj The new version of fast-azure-storage is published :) Now we can merge this PR

jonasfj:
I'll update npm-shrinkwrap.json and land this in production...

PR #95, opened 2017-02-14
fix spelling errors

djmitche:
As found in taskcluster/taskcluster-cli#100

### DIFF
--- a/schemas/create-client-response.yml
+++ b/schemas/create-client-response.yml
@@ -73,7 +73,7 @@ properties:
   disabled:
     description: |
       If true, this client is disabled and cannot be used.  This usually occurs when the
-      scopes avaialble to the user owning the client no longer satisfy the client.
+      scopes available to the user owning the client no longer satisfy the client.
     type: boolean
 additionalProperties:       false
 required:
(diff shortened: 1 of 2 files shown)


djmitche:
I'll keep pushing to my own repo as a reminder that we should be writing test automation so that it's usable by non-employees :)

jonasfj:
Fair point :)

Den 14. feb. 2017 10.26 AM skrev "Dustin J. Mitchell" <
notifications@github.com>:

petemoore:
@djmitche @jonasfj bizarrely, this change seems [not to have propagated](http://schemas.taskcluster.net/auth/v1/create-client-response.json).

What is even stranger, is it looks like it did propagate at some point, but then regressed, since [this commit](https://github.com/taskcluster/taskcluster-client-go/commit/09445791ef444a8cfba552c325cd8ee88f328386#diff-a2caf70f51a48b706e720c623acfe14bR211) shows that at some point the json schema must have been successfully published. Notice how the diff is the wrong way around, it shows the schema change being reverted, rather than applied.

Is there any automation that could be publishing stale schemas?

djmitche:
Staging was set to publish.. oops!

I fixed that, and I also deployed this change to prod.

PR #97, opened 2017-02-22
Allow azure read-write access to grant read-only as well

imbstack:
This is the simplest approach I could come up with that would grant access how we want and preserve nice error messages. Does it make sense to you?

### DIFF
--- a/src/azure.js
+++ b/src/azure.js
@@ -79,12 +79,10 @@ api.declare({
     });
   }
 
-  // Check that the client is authorized to access given account and table
-  if (!req.satisfies({
-    account:    account,
-    table:      tableName,
-    level:      level,
-  })) {
+  // We have a complicated scope situation for read-only since we want
+  // read-write to grant read-only permissions as well
+  if (!(level === 'read-only' && req.satisfies({account, table: tableName, level: 'read-write'}, true)) &&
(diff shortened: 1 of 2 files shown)


src/azure.js
```diff
@@ -79,13 +79,29 @@ api.declare({
...
-    table:      tableName,
-    level:      level,
-  })) {
-    return;
+  // We have a complicated scope situation for read-only since we want
+  // read-write to grant read-only permissions as well
+  // req.satisfies will throw an error if it is not satisfied
+  if (level === 'read-only') {
+    try {
+      req.satisfies({
```
jonasfj: `req.satisfies` does not throw an exception...

It writes a response and returns `false`, or returns `true`. If called like: `req.satisfies({..}, true)` then it won't write a response...

Thus:
```js
if (!(level == 'read-only' && res.satisfies({account, table, level: 'read-write'}, true) && !res.satisfies({account, table, level})) {
  return;
}
```

should do the trick...

Note, you can also do: `if (!req.satisfies([['scope1'], ['scope2']])) { return; }`, then it's either scope1 or scope2.

imbstack: wait. noooooooo! This means I was wrong all along!

jonasfj: But wasn't it nice of me to not point that out :)

PR #98, opened 2017-03-06
Allow overriding buckets for redeploying

imbstack:
I believe this will work in our current Heroku environment as well, since these will be undefined and the defaults will take over. This is designed to work with https://github.com/imbstack/taskcluster-standalone and it seems to work correctly there. I get docs/schemas/references uploaded to my personal buckets and the service runs happily in Kubernetes.

If there's a better way to go about this, let's do that instead. Also very open to picking different names for these fields.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -49,6 +49,14 @@ defaults:
     statsum:
       secret:                   !env STATSUM_API_SECRET
       baseUrl:                  !env STATSUM_BASE_URL
+
+    # Overrides of s3 buckets for redeployability
+    # Leave unset if running inside Heroku
+    buckets:
+      docs:                     !env DOCS_BUCKET
+      schemas:                  !env SCHEMAS_BUCKET
+      references:               !env REFS_BUCKET
+
   server:
     # Public URL from which the server can be accessed (used for persona)
     publicUrl:                https://auth.taskcluster.net
--- a/src/main.js
+++ b/src/main.js
@@ -93,7 +93,8 @@ let load = Loader({
     requires: ['cfg'],
     setup: ({cfg}) => Validate({
       prefix:  'auth/v1/',
-      aws:      cfg.aws
+      aws:      cfg.aws,
+      bucket:   cfg.app.buckets.schemas,
     })
   },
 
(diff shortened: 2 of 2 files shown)


jonasfj:
In short, I would like this for tc-lib-docs. But let's not persist the other publication means any further.
For those we either publish the way we do now, or disable publishing? And in future when we get it clean up we just disable it, I hope :)

djmitche:
I'm not sure what Jonas's comments mean, but this looks good to me.

I suspect Jonas is saying that other installs should still use our schemas / references.  I don't think that makes much sense, as those installs may not be running the latest versions of all of the services.

jonasfj:
We currently publish references/schemas in two ways:
 1) through tc-lib-docs (the new way)
 2) through tc-lib-api and tc-validator, and pulse-publisher (the old way)

I suggest that we don't persist (2), and instead modify (1) so that it can fully replace everything that happens in (2). I think we talked about this when we did tc-lib-docs, we just never finished moving away from the old form of publishing schemas and references...

What I'm saying is that we shouldn't complicate our code with logic to keep up (2), instead we should migrate to only publish docs using tc-lib-docs, this would greatly simplify our lives.

jonasfj:
> I suspect Jonas is saying that other installs should still use our schemas / references. I don't think that makes much sense, as those installs may not be running the latest versions of all of the services.

No, I'm saying installs should have their own docs site... But that we don't want each install to have a docs site, a schemas bucket and a references bucket. We should publish schemas and references as part of the docs site, so we don't need the schemas bucket and references bucket.

Then we only need a docs site for each deployment of taskcluster.

For this to work well, we probably need to make the `id`s of the schemas relative links... So that the docs site can make them non-relative when it publishes them under a different hostname than `schemas.taskcluster.net`.

In fact we should modify the docs site we have now such that schemas a published at something like: `docs.taskcluster.net/<service>/schemas/v1/<filename>.json` and references at something like: `docs.taskcluster.net/<service>/references/v1/(api|events).json`.

djmitche:
This would be a pretty big change.  When I was looking to replace http with https in schemas/references, it turned out to be too big a hairball.  

@imbstack's change is simple enough, and removes an obstacle to redeployment.  My vote is to merge this, and record the proposed combination of schemas, references, and docs in a new bug.  Note that @jhford today expressed a desire to not automatically create clients for services like host-secrets, so there is probably still a need for some customizability.

PR #100, opened 2017-03-16
use res.reportError instead of res.status

djmitche:


djmitche:
Staging is failing to deploy..

```
Error: Cannot find module './build/Debug/buffertools.node'
```

djmitche:
OK, brian gave me the magic turn-off-caching env:
```
heroku config:set NODE_MODULES_CACHE=false --app taskcluster-auth-staging
```
and I then pushed a whitespace change to rebuild the buildpack, ran checkStaging, and promoted.

PR #102, opened 2017-03-31
Add audit log

imbstack:
This is pretty straightforward I guess. Biggest question would be if what is being logged makes sense.

I'm trying this out on staging now and it seems to work. It is possible that won't hold up at production levels of workload though, so I'll keep a close eye on this when I roll it out.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -57,6 +57,9 @@ defaults:
       schemas:                  !env SCHEMAS_BUCKET
       references:               !env REFS_BUCKET
 
+    # Name of a kinesis firehose stream to write audit logs to
+    auditLog:                   !env AUDIT_LOG
+
   server:
     # Public URL from which the server can be accessed (used for persona)
     publicUrl:                https://auth.taskcluster.net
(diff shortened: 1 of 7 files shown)


imbstack:
> That's definitely going to be a lot of traffic - something like 250rps! Is that going to end up costing Real Money?

> A few things come up in looking at this:

>    It might be good to make this only occur if auditLog is set, so then disabling is as simple as changing a setting
    It's probably worth removing the console.log statement now. This is in the critical path for every API request, so every bit of savings counts.
    Looking at papertrail it seems that this is
        double-logging everything (once for authenticateHawk, once for the actual request). We could certainly save by not logging authenticateHawk, and probably further by not even verifying the signature on the authenticateHawk call (it requires no scopes)
        not logging the correct clientId, probably due to use of a too-old tc-lib-api? project/taskcluster/aws-provisioner/production and project/taskcluster/tc-queue/production are both the issuers of the temporary certificates actually doing the work, and those temporary certs' clientIds are of far more interest.
    taskcluster-lib-api logs the actual authorization decision. A log message from tc-auth, from this patch, will only mean that the credentials were OK, not that access was actually allowed, which may be confusing or not helpful in a situation where we need to review the logs ("ok, so the attacker managed to call the fooBar endpoint.. well, they sent good credentials, we don't know if access was actually granted")

Depending on cost, it might make the most sense to log both the authentication and the authorization decisions (and, in fact, to log the failures, too).

PR #103, opened 2017-04-12
Use azure-entities 2.0.0

imbstack:
This is the only repo I could find that would not work with the 2.0.0 version of azure-entities which enforced credentials being passed. This upgrades the version of the library and updates the code to work with it. It also removes some old broken bits that broke tests when no azure creds are specified.

### DIFF
--- a/test/helper.js
+++ b/test/helper.js
@@ -51,15 +51,15 @@ mocha.before(async () => {
     helper.Client = overwrites['Client'] = data.Client.setup({
       table: 'Client',
       account: 'inMemory',
+      credentials: null,
       cryptoKey,
       signingKey,
-      context: {resolver: overwrites.resolver},
     });
     helper.Role = overwrites['Role'] = data.Role.setup({
       table: 'Role',
       account: 'inMemory',
+      credentials: null,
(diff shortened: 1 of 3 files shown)


imbstack:
I'm having trouble reproducing the travis failure locally. Feel free to hold off on review until I figure it out.

imbstack:
Ok, just had to kick caches and rerun to make them pass :/

Feel free to review whenever!

djmitche:
That's OK, SpaceX has a guy whose job it is to kick any Merlin engines that don't start the first time.  A good whack with a spanner (and "engineer's persuader") always does the trick.

test/helper.js
```diff
@@ -51,15 +51,15 @@ mocha.before(async () => {
     helper.Client = overwrites['Client'] = data.Client.setup({
       table: 'Client',
       account: 'inMemory',
+      credentials: null,
       cryptoKey,
       signingKey,
-      context: {resolver: overwrites.resolver},
```
djmitche: This appears to be passed in to the functions that need it now (just noting for my reference..)

PR #107, opened 2017-06-30
Added webhooktunnel endpoint

ckousik:
Added an API endpoint for obtaining a token and ID for connecting to webhooktunnel.

### DIFF
--- a/schemas/webhooktunnel-token-response.yml
+++ b/schemas/webhooktunnel-token-response.yml
@@ -1,0 +1,20 @@
+$schema:  http://json-schema.org/draft-04/schema#
+title:                      "Webhooktunnel Token Response"
+description: |
+    Token for connecting a worker to webhooktunnel proxy
+type:                       object
+properties:
+  id:
+    type: string
+    title: Proxy Connection ID
+    description: |
+      id for proxy connection
+  token:
+    type: string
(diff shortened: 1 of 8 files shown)


schemas/webhooktunnel-token-response.yml
```diff
@@ -0,0 +1,20 @@
...
+properties:
+  id:
+    type: string
+    title: Proxy Connection ID
+    description: |
+      id for proxy connection
+  token:
+    type: string
+    title: Auth Token
+    description: |
```
ckousik: wouldn't decoding and jwt verification fail if it is not properly base64 encoded?

jonasfj: sure, it could cause issues elsewhere...

Actually, you're perfectly right that it's not necessary... you can leave it out if you want to. In fact it might be overkill to add a regex for jwts... I just generally to be as explicit as reasonably possible in the JSON schemas, as it catches a lot of bugs..

src/main.js
```diff
@@ -48,7 +48,7 @@ let load = Loader({
             expires:  taskcluster.fromNowJSON('24 hours'),
           };
 
-        },
+        },
```
jonasfj: @ckousik, still missing this one :)

PR #110, opened 2017-09-27
Do not attempt to return an emtpy json response when deleting a role

gregarndt:
Looking at the other delete method we have, it seems like this might do the trick? (cc @djmitche @eliperelman )

### DIFF
--- a/src/v1.js
+++ b/src/v1.js
@@ -719,7 +719,7 @@ api.declare({
     this.resolver.reloadRole(roleId)
   ]);
 
-  return res.reply();
+  return res.status(200).send();
 });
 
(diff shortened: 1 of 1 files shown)


gregarndt:
Ah darn it! I was going to do that at first too but knew deleting a client ID worked, so I just duplicated it there.

eliperelman:
`DELETE`s are supposed to be idempotent, so I'm not sure why we would need any kind of response from a `DELETE`. Wouldn't it be more correct to make this not return a JSON response?

eliperelman:
Yep, and 202 also works if the work has been queued but not completed.

PR #111, opened 2017-11-14
Relax about engines

djmitche:
Just so I can run `yarn test` without it barfing on me.


djmitche:
sure, that's fine too

PR #112, opened 2017-11-15
Spring cleaning

imbstack:
The `--exit` mocha opt I added is explained in the commit message to d374a5b. If we would rather make this work better than just patch over an issue, I'm happy to go about making that happen as well.


package.json
```diff
@@ -15,33 +15,33 @@
     "url": "https://github.com/taskcluster/taskcluster-auth.git"
   },
   "dependencies": {
-    "assume": "^1.3.1",
-    "aws-sdk": "^2.149.0",
```
djmitche: in the time between the last commit and this one, they ugpraded? :)

imbstack: Wow... looks like they did!

djmitche:
Let's move fast and break things!!

djmitche:
Reminder to self to merge https://github.com/taskcluster/taskcluster-docs/pull/225 once this lands.

djmitche:
🍿

PR #113, opened 2017-11-22
General Cleanup

djmitche:
This is in preparation for rfc48.  It's based on spring-cleaning, but I can rebase on master once that lands.

This mostly addresses tests, but also puts some docs in docs/.  I'll make a PR in tc-docs to remove the relevant content there (and leave a more general approach).


djmitche:
Good point :)

PR #114, opened 2017-11-22
Parameterized Roles

djmitche:
I think this is about 80% ready.  It's feature-complete, but I need to see what the performance looks like in real life and consider adding some caching at the resolver level.

The commits have long comments on them, so please review commit-by-commit.


src/scoperesolver.js
```diff
@@ -237,34 +230,156 @@ class ScopeResolver extends events.EventEmitter {
...
+    // encode the roles as rules, including the `assume:` prefix, and marking up
+    // the expansions of any parameterized scopes as {scope, index}, where index
+    // is the index in the input at which the replacement begins (the index of
+    // the `*`)
+
+    let rules = [];
+    roles.forEach(({roleId, scopes}) => {
+      let pattern = `assume:${roleId}`;
+      if (pattern.endsWith('*')) {
+        const index = pattern.length - 1;
```
jonasfj: Still not convinced this makes any sense... you have the index as you traverse the trie...

src/scoperesolver.js
```diff
@@ -237,34 +230,156 @@ class ScopeResolver extends events.EventEmitter {
...
+        rules.push({pattern, scopes: scopes.map(scope =>
+          PARAMETER.test(scope) ?  {scope, index} : scope)});
```
jonasfj: Why care, why not just always make the substitution... if the rule matched ends with `*`

src/scoperesolver.js
```diff
@@ -237,34 +230,156 @@ class ScopeResolver extends events.EventEmitter {
...
+  buildResolver(roles) {
+    ScopeResolver.cycleCheck(roles);
+
+    // encode the roles as rules, including the `assume:` prefix, and marking up
+    // the expansions of any parameterized scopes as {scope, index}, where index
+    // is the index in the input at which the replacement begins (the index of
+    // the `*`)
+
+    let rules = [];
+    roles.forEach(({roleId, scopes}) => {
```
jonasfj: nit: Wouldn't `.map` be nicer...

src/scoperesolver.js
```diff
@@ -237,34 +230,156 @@ class ScopeResolver extends events.EventEmitter {
...
+              return expandedScope.replace(PARAMETER_TO_END, scope.slice(index));
+            } else {
+              return expandedScope.replace(PARAMETER_G, scope.slice(index));
+            }
+          }));
+      }
+
+      // TODO: normalize should remove equivalent scopes too
+      // https://github.com/taskcluster/taskcluster-lib-scopes/pull/17
+      scopes = [...new Set(scopes)];
```
jonasfj: If instead of using a `Set` for `seen` you used a sorted list from `normalizeScopeSet` you could do a binary search to see if it contains any scope that is greater than the given scope. In fact it would be fairly cheap (similar to a merge operation) to find the scopes from `executeTrie(dfa, scope).map(...` that are not covered by another scope in `seen` and add those to `scopes`.

Then you would instantiate `seen = normalizeScopeSet(input.sort(scopeCompare))` and `scopes = [...seen]`, obviously skipping the `if (seen.has(scope)) { continue`.

This would fix the case where `assume:*` takes a look time to expand, because `assume:*` covers all other `assume:<something>` scopes, hence, they wouldn't need to be added to `scopes`.

src/trie.js
```diff
@@ -0,0 +1,97 @@
+const _ = require('lodash');
+const {scopeCompare} = require('taskcluster-lib-scopes');
+
+/**
+ * Build a trie to expand patterns involving kleene stars.
+ *
+ * Given input of the form [{pattern, scope}], this builds a trie
```
jonasfj: Input is on the form `[{pattern, scopes}]` right?

src/trie.js
```diff
@@ -0,0 +1,97 @@
...
+    let j = i;
+    while (j < n) {
+      // find the current segment, at indices [seg .. j-1]
+      let seg = j;
+      let current = rules[j].pattern[k]; // current character
+      if (current === '*' && rules[j].pattern.length === k + 1) {
+        // segment type A, terminal * in pattern
+        do {
+          j++;
+        } while (j < n && rules[j].pattern[k] == '*' && rules[j].pattern.length === k + 1);
```
jonasfj: This part is necessary if you have multiple rules on the form `{pattern, scope}` where `pattern` is the same string.
You cannot possibly have that.

1. it's not possible with how we define roles
2. if we did have this we should just merge them into a single `{pattern, scopes}` rule, by merging `scopes`
3. this is never generated form `buildResolver`

I assume this is a left over from when you had `{pattern, scope}` instead of `{pattern, scopes}`. We can get some simplicity of case (A) and (B), but not doing this... simply requiring pattern uniqueness for input.

src/trie.js
```diff
@@ -0,0 +1,97 @@
...
+        state.scopes = _.flatten(rules.slice(seg, j).map(r => r.scopes));
```
jonasfj: If you were actually merging anything here you should normalize...

src/trie.js
```diff
@@ -0,0 +1,97 @@
...
+          j++;
+        } while (j < n && rules[j].pattern[k] === current);
+        state[current] = gen(seg, j, k+1);
+      }
+    }
+
+    // a terminal star in the input here (state['*']['end']) will match all
+    // patterns
+    let allScopes = _.flatten(rules.slice(i, n).map(r => r.scopes));
+    if (state['*']) {
```
jonasfj:
```js
state['*'] = state['*'] || {};
state['*'].end = {scopes: allScopes};
```

src/trie.js
```diff
@@ -0,0 +1,97 @@
...
+ * Executes the trie for a single input, returning the accumulated scopes.
+ */
+exports.executeTrie = (trie, input) => {
+  let state = trie;
+  let result = [];
+  let k = 0;
+
+  while (state) {
+    if (state.scopes) {
+      result = result.concat(state.scopes);
```
jonasfj: If instead you did: `result.push({matchedAt: k, scopes: state.scopes})`
Then you won't need to have scopes containing `<..>` to be on the form `{index, scope}`.

Instead when you do `executeTrie` just:
```js
let parameterized = [];
executeTrie(trie, inputScope).map(({matchedAt, scopes}) => {
  if (matchedAt !== null) {
    parameterized = parameterized.concat(scopes.replace('<..>', inputScope.substr(matchedAt)));
  } else {
    parameterized = parameterized.concat(scopes);
  }
});
```

src/data.js
```diff
@@ -195,3 +195,54 @@ Role.prototype.json = function(resolver) {
 
 // Export Role
 exports.Role = Role;
+
+/**
+ * AtomicRoles represents the entire role configuration in one row, to support
+ * atomic updates.
+ */
+var AtomicRoles = Entity.configure({
```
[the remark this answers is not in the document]
djmitche: Oh, I thought it was 1MB :(

src/scoperesolver.js
djmitche: It's true, but with this structure I can cleanly separate the trie from the parameterization.  I didn't see any advantage to mixing the two.

src/scoperesolver.js
djmitche: efficiency.. saving a call to replace which is likely to allocate a new string..

src/data.js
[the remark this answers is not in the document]
djmitche: Suggestions on how to solve this?  I don't think switching to postgres is a good project to tack onto the end of this one.  Given those options, I'd prefer to keep the existing storage -- while it's theoretically racy, in practice roles are only occasionally updated, and it would be easy to set up the ScopeResolver to keep using the old resolver whenever a new resolver contains a cycle.

src/trie.js
djmitche: I rather like the cleanliness of the trie implementation here.  This change requires that the trie implementation know something about parameterization *and* that the caller know something about parameterization.

I don't think what you've suggested would be noticeably more efficient, would it?  If not, I'd like to keep the cleaner separation.

djmitche: Ah, I see that this is useful for other purposes, too -- in particular, being able to sort and normalize `scopes` in the trie.

djmitche: Although it would have to be `{scopes: [{matchedAt, scope}]}` since each can match at a different position in the input.

djmitche: And indeed, the same scope may appear several times with different matchedAt:

```
x* --> b<..>c
xy* --> b<..>c
```
`executeTrie(true, 'xyz') -> `[{scope: `b<..>c', matchedAt: 1}, {scope: 'b<..>c', matchedAt: 2}]`.

djmitche: I should stop writing comments until I figure this out.

src/data.js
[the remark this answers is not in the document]
jonasfj: Okay, it's actually 256kb, let's see how close we are to that...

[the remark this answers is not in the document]
djmitche: From discussion, the race is in fact a vulnerability -- someone could maliciously create a role.  And refusing to adopt a cyclical resolver means that when the dynos restart daily, they will fail to start -- no good.  So options are
 * keep with azure tables if the current size is substantially smaller than [256k](https://github.com/taskcluster/azure-entities/blob/master/src/entitytypes.js#L489)
 * azure blobs (which do have sufficient semantics)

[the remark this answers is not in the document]
djmitche: The current set of roles are 245k.  So I'd say, no, we don't have room.

djmitche:
OK, I haven't touched the storage stuff yet -- I'll come back to that tomorrow.

I modified the trie execution to return the location at which each scope matched, and then extended the resolver to use that information for its parameter substitutions.

On your advice, I've also modified the resolver to keep a normalized `seen` scopeset and use the work queue to flatten the recursion only for scopes that are not already satisfied by `seen`.  This has a big impact on massive expansions that end up reducing to `*` or `assume:*`: once the `*` scope is seen, nothing else goes on the queue.  Times in my testing are down to <10ms for all of the "real roles" performance tests.  It's 500-1000x slower for some of the pathological examples, which is about where the old solution was.

One additional optimization I didn't introduce here: the `see` function currently starts at the beginning of `seen` on every iteration.  However it's quite likely to find its target at the final `i` or `i+1` from the last call, so it probably makes sense to start there and "loop around" to the beginning of `seen` if it does not find the right location immediately.  But there are lots of edge cases that I didn't want to think too hard about.

djmitche:
@jonasfj do you want to look at the storage implementation?

I elected to just vendor azure-blob-storage for now, pending resolution of its package status on npmjs.com.

src/v1.js
```diff
@@ -567,55 +575,66 @@ api.declare({
...
+        ScopeResolver.cycleCheck(checkRoles);
+      } catch (e) {
+        throw reportError('InputError', `Invalid roles: ${e.message}`, {});
+      }
 
-    // If stored role different or older than 15 min we return 409
-    let created = new Date(role.details.created).getTime();
-    if (role.description !== input.description ||
-        !_.isEqual(role.scopes, input.scopes) ||
-        role > Date.now() - 15 * 60 * 1000) {
```
djmitche: note that the old code here compared the role object to a Date object..

test/helper.js
```diff
@@ -125,4 +146,22 @@ mocha.after(async () => {
...
+    try {
+      await blobService.deleteContainer(helper.containerName);
+    } catch (e) {
+      if (e.code !== 'ResourceNotFound') {
+        throw e;
+      }
+      // already deleted, so nothing to do
+      // NOTE: really, this doesn't work -- the container doesn't register as existing
+      // before the tests are complete, so we "leak" containers despite this effort to
+      // clean them up.
```
djmitche: This bums me out, but seems to be how Azure works.  I noticed that deleting blobs has some synchronization issues, too: a DELETE followed immediately by a GET will return the deleted data.

test/role_test.js
```diff
@@ -68,11 +74,15 @@ suite('api (roles)', function() {
...
+
+    await helper.events.waitFor('e1');
+
     assume(role.description).equals('test prefix role');
     assume(new Date(role.created).getTime()).is.atmost(Date.now());
     assume(role.scopes).deep.equals(['dummy-scope-2']);
     assume(role.expandedScopes).contains('dummy-scope-2');
-    assume(role.expandedScopes).does.not.contain('assume:' + roleId);
+    // expandedScopes should always include itself
+    assume(role.expandedScopes).contains('assume:' + roleId);
```
djmitche: I suspect that this used to pass due to a race.  *Everything* else consistently tests that role.expandedScopes contains `assume:${role.roleId}`.

src/v1.js
jonasfj: nice catch

jonasfj:
> I elected to just vendor azure-blob-storage for now, pending resolution of its package status on npmjs.com.


Instead of vendoring you can just reference the github repository by commit hash in `package.json`. See: 
https://docs.npmjs.com/files/package.json#github-urls

In `dependencies` you can have: `"mypackage": "myorg/myrepo#<revision-hash>"`

djmitche:
vendoring via github doesn't work for babel-compiled libraries :(

jonasfj:
oh, that's sad...

--
Regards Jonas Finnemann Jensen.

2017-12-04 12:22 GMT-08:00 Dustin J. Mitchell <notifications@github.com>:

djmitche:
I deployed this to staging, merging f0360f3 to master.  But it doesn't seem to be expanding roles correctly:
```
> staging.role('repo:hg.mozilla.org/try:cron:nightly-*').then(console.log).catch(console.error);
Promise { then: [Function] }
> { roleId: 'repo:hg.mozilla.org/try:cron:nightly-*',
  description: '*DO NOT EDIT*\n\nScopes for nighlty cron tasks triggered from pushes to https://hg.mozilla.org/try\n\nThis role is configured automatically by [taskcluster-admin](https://github.com/taskcluster/taskcluster-admin).',
  created: '2017-12-06T20:47:32.247Z',
  lastModified: '2017-12-06T20:47:32.247Z',
  scopes: [ 'assume:project:releng:nightly:level-1:try' ],
  expandedScopes: [ 'assume:project:releng:nightly:level-1:try' ] }
```

basically, the resolver is acting like the identity function.  I need to investigate this.

Role modifications are frozen in both production and staging.  Nothing is deployed to production yet.  Don't promote staging to production!!

djmitche:
Oh.. that is, in fact, the expansion of that role.

djmitche:
OK, I'm happy with this in staging now.  I'll deploy to production tomorrow when I have time to monitor the load.  Roles will stay locked until then.

djmitche:
screenshot from 2017-12-07 08-36-48

At least from Heroku's metrics, this doesn't seem to have had much impact.  This is without any *actual* parameterized roles, but that shouldn't make any difference.

PR #116, opened 2017-12-15
Escape regex in YAML

jonasfj:
This might be wrong other places too...

Not sure if there is a good way to fix / avoid problems like this...

It's not critical, but it's sure is ugly...


imbstack:
Closing since wontfix and it keeps tricking me into looking at it and thinking that I have an unread review 😮

PR #117, opened 2017-12-20
Allow clock skew in temp creds

jonasfj:
credits to @petemoore for finding this...


petemoore:
We should document that the auth service adjusts for clock skew, so that client libraries, or callers of client libraries, don't also apply clock skew.

And we should probably apply the same clock skew to hawk authentication ([currently we use 15 mins](https://github.com/taskcluster/taskcluster-auth/blob/4ece45d1ab4f5e3e340bb7340e377d84ff16f210/src/signaturevalidator.js#L326-L329), so this should probably be reduced to 5 mins and documented too).

PR #120, opened 2018-01-09
Use morgan

djmitche:
As @jhford suggested -- this will log each request on the webhead that performs it, as well as the time spent.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -51,7 +51,7 @@ let load = Loader({
 
         },
         sentryDSN: async (project) => {
-          let key = await this.sentryManager.projectDSN(project);
+          let key = await sentryManager.projectDSN(project);
           return {
             project,
             dsn: _.pick(key.dsn, ['secret', 'public']),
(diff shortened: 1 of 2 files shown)


src/main.js
```diff
@@ -51,7 +52,7 @@ let load = Loader({
 
         },
         sentryDSN: async (project) => {
-          let key = await this.sentryManager.projectDSN(project);
+          let key = await sentryManager.projectDSN(project);
```
imbstack: I thought we made this change yesterday?

imbstack:
Also where are the tests results O_o

src/main.js
djmitche: Oh no, apparently I didn't land that!  I pushed to origin/master :(

djmitche:
Re test results: EALREADYHAVE99PROBLEMS

djmitche:
I made a followup to put the app.use(morgan..) before the app.use('/v1', ..).

PR #123, opened 2018-01-19
Bug 1431750 - add additionalProperties to schemas where missing

petemoore:
Again, we should wait for the CI to run before merging this! 😉


schemas/authenticate-hawk-response.yml
```diff
@@ -2,8 +2,7 @@ $schema:  http://json-schema.org/draft-06/schema#
 title:                    "Hawk Signature Authentication Response"
 description: |
   Response from a request to authenticate a hawk request.
-type:                     object
```
jonasfj: I have mixed feelings about this one... I can't remember which way I decided to go on the subject...

PR #128, opened 2018-01-28
Use POST for expandScopes

djmitche:
Docker-worker calls the existing (GET-based) expandScopes, so that
remains in place as expandScopesGet, but marked deprecated.  When
docker-worker is upgraded to not use the old method, or is itself
deprecated, we can remove expandScopesOld.

Thanks to @Biboswan for finding this issue in https://github.com/taskcluster/taskcluster-tools/issues/278.

### DIFF
--- a/src/v1.js
+++ b/src/v1.js
@@ -793,6 +793,26 @@ api.declare({
 api.declare({
   method:     'get',
   route:      '/scopes/expand',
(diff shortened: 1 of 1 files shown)


djmitche:
Well, docker-worker calls this a few times a minute, so it works :)

PR #130, opened 2018-01-30
Bug 1434369 - don't try to create tables / containers every time

djmitche:
This isn't the most sophisticated way to do this, but I think it should be adequate..

### DIFF
--- a/src/azure.js
+++ b/src/azure.js
@@ -2,6 +2,13 @@ var _           = require('lodash');
 var azure       = require('fast-azure-storage');
 var api         = require('./v1');
 
+// keyed by account/tableName, the last time createTable was called for the
+// given table.  This is used to avoid lots of redundant calls to createTable
+// for the same table.
+var tableLastCreated = {};
+// Similar for containers
+var containerLastCreated = {};
+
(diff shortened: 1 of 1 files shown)


djmitche:
@jonasfj when you do review this, please deploy it too.  It should cut down on the error rate in auth..

jonasfj:
A neat class with a method to wrap a function could also be nice...
```js
const createTableCache = new Cache();

...
  await createTableCache(key, async () => {
     azure.createTable(key);
  });
```

Ideally we could then create `const createTableCache = new Cache();` on the `context` object in `main.js`...
But this is also fine, if tests pass I doubt there are any bad side-effects...

djmitche:
I like that better, and it's a bit more testable.

djmitche:
Honestly, that kind of turns this into a big deal -- suddenly main.js has to create two objects whose only purpose is to create Azure tables, etc.  I'll land this as-is.

jonasfj:
Fair point... We can always make this nicer later...

Note. I'm mostly concerned for side effects when testing.. but clearly tests pass here..

PR #132, opened 2018-02-01
Switch lru package to fix osx compilation

imbstack:
I'm not sure what will happen perforamance-wise, but this one seems to work ok.


imbstack:
Oops, also added my lib-monitor update in here, but it should be fine.

PR #135, opened 2018-02-06
Bug 1436112 - Rename blob to container

djmitche:
I'm not *too* concerned about compatibility here -- the only thing that uses azureBlobSAS is tc-backup.  AWS provisioner used to, but does no longer.

If there *were* other uses of this method, they would
 * fail immediately unless scopes were updated, as the required scopes have changed (I did modify all scopes I found in the scope explorer, which were tc-backup, aws-provisioner-staging, and some jungle stuff)
 * get errors on updating taskcluster-client*, as the method has gone away.


jonasfj:
Yay, I managed to land mine first :)

Sorry for breaking your changes.. But scope-expressions might cause a conflict here.

---
On topic: this seems really decent... I love better naming.

PR #137, opened 2018-02-13
Add bang symbol to the client ID schema

helfi92:
Auth0 doesn't provide a nice user ID for GitHub logins. For GitHub, we
receive IDs of the form `github|12334`. For better user IDs, we append the
GitHub nickname to the ID (e.g, `github|1234/helfi92`). Encoding the
latter gives `github%7C1234%2Fhelfi92`. To be able to use this,` %` needs to
be included in the Client ID schema.

See also:
https://github.com/taskcluster/taskcluster-login/pull/71#issuecomment-364785926
https://github.com/taskcluster/taskcluster-login/pull/71#discussion_r167710957

### DIFF
--- a/schemas/constants.yml
+++ b/schemas/constants.yml
@@ -5,7 +5,7 @@ access-token-pattern:     "^[a-zA-Z0-9_-]{22,66}$"
 # Printable ascii string for roleId
 roleId: "^[\\x20-\\x7e]+$"
 
-clientId: "^[A-Za-z0-9%@/:.+|_-]+$"
+clientId: "^[A-Za-z0-9!@/:.+|_-]+$"
 
 # Slugid pattern, for when-ever that is useful
 slugid-pattern: "^[A-Za-z0-9_-]{8}[Q-T][A-Za-z0-9_-][CGKOSWaeimquy26-][A-Za-z0-9_-]{10}[AQgw]$"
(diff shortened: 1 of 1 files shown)


jonasfj:
hmm, I'm starting to question the brilliance of `%` encoding...

Notably, there is a lot of risk that if you don't double encode it when inserting it into an API end-point... then it'll be decoded.

@helfi92, maybe it's better to use a different encoding? ie. I'm suggesting we do the same as percent-encoding except we use a different special character than `%`.

Example we could do `encodeURIComponent(myString.replace(/[!]/g, '!21')).replace(/[%]/g, '!')`

So:
 1. Replace `!` with `!21`
 2. Encoding with `encodeURIComponent` (which won't encode `!`)
 3. Replace `%` with `!`

decoding is then: `decodeURIComponent(myEncodedString.replace(/[!]/g, '%'))`
Notably:
 1. Replace `!` with `%`
 2. Decode with `decodeURIComponent` (which converts `%21` to `!`)

Also, instead of using `github|1234/helfi92` which would cause `/` to be encoded... how about we just append `|<githubUserName>` so you get: `github|1234|helfi92`

Whether we use slash or pipe doesn't really matter since we're always appending it no matter what.

helfi92:
The change to the encoding and decoding is introduced [here](https://github.com/taskcluster/taskcluster-login/pull/71/commits/7b0e423ef9eb58b25c0b6ffec684fabcb4bb8e74).

djmitche:
I think we don't need this change anymore, then?

helfi92:
> I think we don't need this change anymore, then?

No, but I think we still need to add the `!` symbol.

helfi92:
Will merge this tomorrow.

PR #138, opened 2018-02-28
Complete trie

jonasfj:
See: `trie.js` and `scopesetbuilder.js` those are interesting.

In `trie_test.js` please review the `evalRulesRecursively` method as it is important..
The rest you can sort of scan...


src/scoperesolver.js
```diff
@@ -236,192 +243,60 @@ class ScopeResolver extends events.EventEmitter {
...
-      return mergeScopeSets(inputs, seen);
+      return ScopeSetBuilder.mergeScopeSets(scopes, inputs);
     };
-  };
+  }
 
   /**
    * Return a normalized set of scopes that `scopes` can be expanded to when
    * assuming all authorized roles.
    */
```
djmitche: It looks like the `static normalizeScopes` method is no longer used?

src/scopesetbuilder.js
```diff
@@ -0,0 +1,179 @@
...
+ *
+ * To make it even faster we allow nodes to run a different node (or this) from
+ * the node.next() function, such that the return value is the tree after
+ * consumption of the current value. As a final technicality we leave node.value
+ * undefined until the first call of node.next(), such that we can support empty
+ * nodes like this, by returning null from node.next().
+ */
+class EmptyNode {
+  next() { return null; }
+};
```
djmitche: class appears to be unused..

src/scopesetbuilder.js
```diff
@@ -0,0 +1,179 @@
+const {scopeCompare, mergeScopeSets, normalizeScopeSet} = require('taskcluster-lib-scopes');
+
+/**
+ * A ScopeMergeTree is an object-structure that holds a tree of scope sets as
```
djmitche: *sorted* scopesets

src/scopesetbuilder.js
```diff
@@ -0,0 +1,179 @@
...
+ * array for all intermediate results. The complexity of merging N scopes split
+ * between M scope-sets approaches something like O(N * M).
+ *
+ * Instead we shall create a binary tree of nodes where each leaf holds a
+ * scope-set, then we compute the next value for internal node by taking the
+ * value from a child. This leaves us with an average complexity closer
+ * something like O(N * log(M)), notes this just not proven math, merely a
+ * casual guess backed by benchmarks.
+ *
+ * To make it even faster we allow nodes to run a different node (or this) from
```
djmitche: run -> return

src/trie.js
```diff
@@ -1,102 +1,518 @@
...
+const PARAM = '\u001a';
+const PARAM_TO_END = /\u001a.*$/;
+
+/**
+ * Transform rules from {pattern, scopes} to {pattern, scopes, matched, paramed},
+ * where:
+ *  - matched, is the non-parameterized scopes, and
+ *  - paramed, is the parameterized scopes with PARAM instead of '<..>'
+ */
+const transformRules = (rules) => rules.map(({pattern, scopes}) => {
```
djmitche: Please move this method down near the topological sorting code

src/trie.js
```diff
@@ -1,102 +1,518 @@
...
+  const hasKleene = pattern.endsWith('*');
```
djmitche:
```
if (!hasKleen) {
  return {pattern, scopes, matched: scopes, paramed: []};
}
```

src/trie.js
```diff
@@ -1,102 +1,518 @@
...
+      // We forbid more than one parameter injection
+      if (s.includes('<..>')) {
+        const err = new Error(`parameterized scope '${scope}' contains multiple '<..>'`);
+        err.code = 'InvalidScopeError';
+        err.scope = scope;
+        throw err;
+      }
+      // We forbid ambiguous kleenes
+      if (s.endsWith('*' + PARAM)) {
+        const err = new Error(`parameterized scope '${scope}' ends with '*<..>' which implies an ambiguous kleene`);
```
djmitche: I think this needs to be defined somewhere in the docs..

src/trie.js
```diff
@@ -1,102 +1,518 @@
...
+    this.paramed = paramed;       // scopes given if input traverses through this node with parameter
+    this.kleeneOnly = kleeneOnly; // scopes given in addition to end, enter, paramed if input ends with kleene
+    this._kleeneCache = null;     // scopes given if input ends with kleene at this node (lazily computed)
+    this.children = new Map();    // map from next character to child node
+  }
+
+  dump() {
+    return {
+      end: this.end,
+      enter: ScopeSetBuilder.normalizeScopeSet([...this.enter, ...withParam(this.paramed, '<..>')]),
```
djmitche: Not that it matters outside of tests, but why is paramed not listed separately here?

src/trie.js
```diff
@@ -1,102 +1,518 @@
...
+      // the prefix (or create such a child), and merge node into that...
+      let child = this.children.get(prefix[0]);
+      let isNew = !child;
+      if (isNew) {
+        child = new Node();
+      }
+      modified = child.merge(prefix.slice(1), node);
+      // We only save a new node if merging into it modified it, this prevents
+      // us from poluting the trie with empty nodes.
+      if (modified && isNew) {
```
djmitche: when would we have isNew and not modified?

src/trie.js
```diff
@@ -1,102 +1,518 @@
...
- *   ...
- *   'end': { ... },      // state if this is the end of the input
- * ```
+ * Example, given a trie with children for 'abc' as:
+ *   node  --a-->  A  --b-->  B  --c-->  C
+ * then traverse(node, 'abc', visit) will call visit(A, 'a') and visit(B, 'ab')
+ * and finally return C. Or null, if somewhere along the path to there is no
+ * child for the next character.
+ */
+const traverse = (node, path, visit = () => {}) => {
```
djmitche: Why not make this a method of Node?

src/trie.js
```diff
@@ -1,102 +1,518 @@
...
+ *
+ * r1:   A  -> B
+ * r2:   B  -> C
+ * r1 depends on r2
  *
- * Note that, as applied to roles, the PATTERN for every rule begins with
- * `assume:`, but nothing in this file is specific to that prefix.
+ * r1:   A  -> B
+ * r2:   B* -> C
+ * r1 depends on r2
```
djmitche: Also
```
r1: A -> Bx
r2: B* -> C
```

src/trie.js
```diff
@@ -1,102 +1,518 @@
...
+ *
+ * For any two rules there is a dependency relationship if a parameterization
+ * of one of the rules grants scopes that matches the other scope.
+ *
+ * This returns an topological ordering of rules, where <..> have been replaced
+ * with PARAM in the scopes. If there is a dependency cycle this method will
+ * throw an error with err.code = 'DependencyCycleError' and an err.cycle
+ * property, this is the only effect of this method that useful outside this
+ * file (and testing).
```
djmitche: This should describe InvalidScopeError too.

src/trie.js
```diff
@@ -1,102 +1,518 @@
...
-        } while (j < n && rules[j].pattern[k] === current);
-        state[current] = gen(seg, j, k+1);
-      }
+const dependencyOrdering = (rules = []) => {
+  rules = transformRules(rules);
+  // For each rule we must have an efficient way to find other rules that it
+  // depends on. To do this we build a trie (because tries are fast), where
+  // matching a rule gives the scope of its index value in the rules array.
+  // This is a bit of a hack, but so long as numbers don't end with '*' the
+  // ScopeSet will not merge them (which would be bad).
```
djmitche: This is kind of gross, but it works :)

src/trie.js
```diff
@@ -1,102 +1,518 @@
...
+  const ordering = [];    // Final topological ordering
+  const seen = new Set(); // rules we've seen (ie. this is the stack)
+  const done = new Set(); // rules we've already output to the ordering
+
+  // visit a rule to add it to the ordering
+  const visit = (R) => {
+    if (done.has(R)) {
+      return; // Skip, if already in the ordering
+    }
+    // If R is on the stack, we have a cycle...
```
djmitche: .. is on the stack, but not done, then ..

src/trie.js
```diff
@@ -1,102 +1,518 @@
...
+ *   execute(node, prefix + input)
+ *
+ * This method doesn't not interpret '*' at the end of prefix as kleene. Hence,
+ * the above equivalence does not hold for prefix = '...*' and input = ''.
+ * In practice we forbid rules on the form 'A*<..>' because '*' is ambiguous,
+ * hence, in withPrefix(node, 'A*') we shall never interpret '*' as kleene.
+ *
+ * Notice, this will NOT modify the input trie, but some of its children may be
+ * referenced in the result of this method. Hence, modifying the resulting trie
+ * will invalidate the input trie, and vice versus.
```
djmitche: versus -> versa

src/trie.js
```diff
@@ -1,102 +1,518 @@
...
+  });
+
+  // if there is a target we have to add it's scopes too
+  let end = [];
+  let kleeneOnly = [];
+  if (target) {
+    end = target.end;
+    kleeneOnly = target.kleeneOnly;
+    enter.add(target.enter);
+    paramed.add(target.paramed); // <..> prefixed with empty remainder
```
djmitche: This comment baffled me until I worked it out myself.  How about `// equivalent to withParam(target.paramed, '' + PARAM)`

src/trie.js
```diff
@@ -1,102 +1,518 @@
...
+      // of the parameter, s.t. scope == prefix + PARAM + suffix
+      const [prefix, suffix] = scope.split(PARAM);
+      // We then transform the full existing trie, with the assumption input is
+      // prefix and suffixed as given in the scope, and merge with the current node
+      // Note: because we're inserting in dependency order this cannot possibly depend
+      //       on another rule that haven't been inserted yet, not even itself.
+      let transformed = trie;
+      if (prefix !== '') { // skip withPrefix if there is no prefix
+        // Note that rules on the form: 'A*<..>B' are forbidden, so we need not
+        // worry about the limitations of withPrefix wrt. to input on the form
```
djmitche: remove "to"

src/scoperesolver.js
```diff
@@ -5,14 +5,21 @@ var events      = require('events');
 var LRU         = require('quick-lru');
 var debug       = require('debug')('auth:ScopeResolver');
 var {scopeCompare, mergeScopeSets, normalizeScopeSet} = require('taskcluster-lib-scopes');
```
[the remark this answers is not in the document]
jonasfj: I think we should do in a separate PR, that's all..

src/v1.js
```diff
@@ -573,54 +573,47 @@ api.declare({
...
-  let when = new Date().toJSON();
-  role = {
+  let role = {
     roleId,
     description: input.description,
     scopes: input.scopes,
-    lastModified: when,
-    created: when,
+    lastModified: new Date().toJSON(),
+    created: new Date().toJSON(),
```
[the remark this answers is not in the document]
jonasfj: No reason... I should probably make it back :)
I just didn't a bunch of refactor for all of this...

package.json
```diff
@@ -48,6 +49,7 @@
     "typed-env-config": "^1.1.0"
   },
   "devDependencies": {
+    "0x": "^3.3.0",
```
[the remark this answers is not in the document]
jonasfj: oh, yeah, I should remove that, thanks

src/scoperesolver.js
jonasfj: I'm not sure it was used before... there is a few things like that. I prefer to cleanup later.

src/scopesetbuilder.js
jonasfj: At one point I had the others derive from this one... I could add that back, just because it's pretty :)

The goal was to document the interface that these nodes must implement.

src/trie.js
jonasfj: I actually meant to remove this... dump() was exclusively for debugging, we can keep it :)

I just wanted to keep a minimalist representation of the nodes...

src/trie.js
jonasfj: If we are merging in an empty node...

`isNew` is true when the child doesn't exist... `modified` is also true if merging the (new child) with the target is modifying the child... 

If we have `isNew == true && modified == false` then we've just merged an empty node into a new empty child node. There is no reason to add that child node, since we can just immediately delete it (it's empty).

I'm not 100% sure this happens. I'm pretty sure it happens a lot in `withSuffix` that we produce empty nodes... I could be wrong, because now most of them probably gets some value in `kleeneOnly`.

jonasfj: I think of it this way: the optimization is safe; if I always apply those then it won't matter if I do simple things else where.

Example: `withPrefix(node, 'no-matches')` will return an empty node.

Instead of testing whether nodes are empty, or if they have sub-trees that are empty, I just always merge them and then this neat optimization ensures that empty children disappear :)

src/trie.js
jonasfj: yeah, before I used a trie.. the topological sorting time was actually dominating the trie generation time.

Especially, for the big cases... It seems crazy, but a naive algorithm has a lot of comparisons to make.

jonasfj:
I'll update docs, thanks for pointing it out.

@djmitche, if you agree with the specification: `execute(withSuffix(trie, suffix), input) == execute(trie, input + suffix)` maybe stop looking at my implementation. Read `Node` and figure out how you would implement `withSuffix` :)

It's sort of a graph/trie problem. And given how clear the definition is, I think we can even prove correctness. But I agree my implementation is not straight forward anymore.

-----
The upside with this approach is that `withPrefix` and `withSuffix` have crystal clear definitions and that we can test extremely well. And if the implementations match up to the definitions, I'm pretty sure what we do in `trie.build` is provably correct.

test/scoperesolver_test.js
```diff
@@ -1,10 +1,12 @@
 suite('scoperesolver', () => {
-  let ScopeResolver = require('../src/scoperesolver');
-  let {mergeScopeSets, scopeCompare} = require('taskcluster-lib-scopes');
-  let Monitor            = require('taskcluster-lib-monitor');
-  let assert = require('assert');
-  let _ = require('lodash');
-  let fs = require('fs');
+  const ScopeResolver = require('../src/scoperesolver');
+  const {mergeScopeSets, scopeCompare} = require('taskcluster-lib-scopes');
+  const Monitor            = require('taskcluster-lib-monitor');
```
imbstack: rid us of this horrid space while we're at it!

jonasfj: fine :)

package.json
```diff
@@ -19,6 +19,7 @@
     "aws-sdk": "^2.150.0",
     "azure-blob-storage": "^2.1.0",
     "azure-entities": "3.0.0",
+    "collections": "^5.1.2",
```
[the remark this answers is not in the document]
jonasfj: nice catch

src/scopesetbuilder.js
```diff
@@ -0,0 +1,179 @@
...
+        this.value = b;
+        this.B = this.B.next();
+      }
+    }
+
+    // If the value selected ends with kleene, we advanced both children until
+    // they are null or we have value that doesn't start with said prefix.
+    if (this.value.endsWith('*')) {
+      const prefix = this.value.slice(0, -1);
+      while (this.A && this.A.value.startsWith(prefix)) {
```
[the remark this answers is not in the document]
jonasfj: `node.value` cannot ever be `null`, except that reading `node.value` is undefined behavior until `node.next()` have been at-least once.

And `node.next()` may return `null` in which case `node.value` is obviously invalid.

We call `A = A.next()` for `A` and `B` in the constructor, so as long as `this.A` is not `null` then `A.value` must be a valid string.

docs/roles.md
```diff
@@ -80,21 +80,12 @@ a scope-set containing `"assume:project-admin:zap"` would expand to include
...
-```
-
-Such cyclical configurations of roles are forbidden and will result in an
-error. In practice, such configurations are always a mistake, anyway.
+### Limits
+ * Parameter substitution `<..>` may only appear once in a scope.
+ * Scopes on the form `prefix*<..>` are not permitted, regardless of what
+   `prefix` is, as the `*` star is ambiguous and leads to unsound expansions.
+ * Roles must be acyclic regardless of the parameter, even `*` as discussed
+   in the next section.
```
djmitche: Some more detail might help here, in particular defining "acyclic'.

djmitche:
@jonasfj do you want to land this?  I think it was basically ready, and seems not to have bitrotted yet..

jonasfj:
yes, this on my TODO list, basically I have to clean up some roles, and that side-tracked...

djmitche:
:tada:

PR #143, opened 2018-03-15
WIP to make test run without pulse credentials

fiennyangeln:
It is done by creating fake helper attributes

Related to Bug 1436799

@djmitche please review,  i tried to run the clienttest without credentials, so I commented the part which have !havePulseCredentials, then I look at each test, what methods they need and try to add it

### DIFF
--- a/src/v1.js
+++ b/src/v1.js
@@ -20,6 +20,7 @@ const roleToJson = (role, context) => _.defaults(
 /** API end-point for version v1/ */
 var api = new API({
   title:      'Authentication API',
+  name:       'auth',
   description: [
     'Authentication related API end-points for Taskcluster and related',
     'services. These API end-points are of interest if you wish to:',
--- a/test/client_test.js
+++ b/test/client_test.js
@@ -10,9 +10,9 @@ suite('api (client)', function() {
   var taskcluster = require('taskcluster-client');
 
   if (!helper.hasPulseCredentials()) {
-    setup(function() {
-      this.skip();
-    });
+    // setup(function() {
+    //   this.skip();
+    // });
   } else {
     const cleanup = async () => {
       // Delete all non-static clients and roles
--- a/test/client_test.js
+++ b/test/client_test.js
@@ -9,19 +9,14 @@ suite('api (client)', function() {
   var testing     = require('taskcluster-lib-testing');
   var taskcluster = require('taskcluster-client');
 
-  if (!helper.hasPulseCredentials()) {
-    // setup(function() {
-    //   this.skip();
-    // });
-  } else {
-    const cleanup = async () => {
-      // Delete all non-static clients and roles
-      await helper.Client.scan({}, {handler: c => c.clientId.startsWith('static/') ? null : c.remove()});
-      await helper.Roles.modify((roles) => roles.splice(0));
-    };
-    setup(cleanup);
-    teardown(cleanup);
-  }
+  const cleanup = async () => {
+    // Delete all clients and roles
+    await helper.Client.scan({}, {handler: c => c.clientId === 'root' ? null : c.remove()});
+    await helper.Roles.modify((roles) => roles.splice(0));
+  };
+  setup(cleanup);
+  teardown(cleanup);
+
 
   test('ping', async () => {
     await helper.auth.ping();
@@ -37,10 +32,6 @@ suite('api (client)', function() {
   });
   const CLIENT_ID = 'nobody/sds:ad_asd/df-sAdSfchsdfsdfs';
   test('auth.deleteClient (non-existent)', async () => {
-    await helper.authEvents.clientDeleted({
-      clientId:  CLIENT_ID,
-    });
-
     await helper.auth.deleteClient(CLIENT_ID);
     await helper.auth.deleteClient(CLIENT_ID);
 
@@ -70,11 +61,6 @@ suite('api (client)', function() {
   });
 
   test('auth.createClient (no scopes)', async () => {
-
-    await helper.authEvents.clientCreated({
-      clientId:  CLIENT_ID,
-    });
-
     var expires = taskcluster.fromNow('1 hour');
     var description = 'Test client...';
     let client = await helper.auth.createClient(CLIENT_ID, {
@@ -133,10 +119,6 @@ suite('api (client)', function() {
   });
 
   test('auth.createClient (with scopes)', async () => {
-    await helper.authEvents.clientCreated({
-      clientId:  CLIENT_ID,
-    });
-
     var expires = taskcluster.fromNow('1 hour');
     var description = 'Test client...';
     var scopes = ['scope1', 'myapi:*'];
@@ -163,14 +145,10 @@ suite('api (client)', function() {
     assume(client2.scopes).not.contains('assume:client-id:' + CLIENT_ID);
     assume(client2.expandedScopes).not.contains('assume:client-id:' + CLIENT_ID);
 
-  assert.deepEqual(helper.publisher.calls, [{method: 'clientCreated', clientId: CLIENT_ID}]);
+    assert.deepEqual(helper.publisher.calls, [{method: 'clientCreated', clientId: CLIENT_ID}]);
   });
 
   const createTestClient = async () => {
-    await helper.authEvents.clientCreated({
-      clientId:  CLIENT_ID,
-    });
-
     var expires = taskcluster.fromNow('1 hour');
     var description = 'Test client...';
     var scopes = ['scope1', 'myapi:*'];
@@ -178,16 +156,12 @@ suite('api (client)', function() {
       expires, description, scopes,
     });
 
-    assert.deepEqual(helper.publisher.calls, [{method: 'clientDeleted', clientId: CLIENT_ID},
-      {method: 'clientDeleted', clientId: CLIENT_ID}]);
+    assert.deepEqual(helper.publisher.calls, [{method: 'clientCreated', clientId: CLIENT_ID}]);
+    helper.publisher.calls = [];
   };
 
   test('auth.resetAccessToken', async () => {
-
-    helper.authEvents.clientUpdated({
-      clientId:  CLIENT_ID,
-    });
-
+    await createTestClient();
     let client = await helper.auth.resetAccessToken(CLIENT_ID);
     assume(new Date(client.lastRotated).getTime())
       .is.greaterThan(new Date(client.lastModified).getTime());
@@ -200,6 +174,7 @@ suite('api (client)', function() {
   });
 
   test('use client', async () => {
+    await createTestClient();
 
     // Fetch client
     let r1 = await helper.auth.client(CLIENT_ID);
@@ -239,10 +214,7 @@ suite('api (client)', function() {
   });
 
   test('auth.updateClient (no scope changes)', async () => {
-
-    helper.authEvents.clientUpdated({
-      clientId:  CLIENT_ID,
-    });
+    await createTestClient();
 
     var expires = new Date();
     let description = 'Different test description...';
@@ -270,10 +242,7 @@ suite('api (client)', function() {
   });
 
   test('auth.updateClient (with scope changes)', async () => {
-
-    helper.authEvents.clientUpdated({
-      clientId:  CLIENT_ID,
-    });
+    await createTestClient();
 
     var expires = new Date();
     let description = 'Third test description...';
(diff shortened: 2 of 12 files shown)


test/helper.js
```diff
@@ -87,7 +97,65 @@ mocha.before(async () => {
...
+    };
+
+    helper.auth = new helper.Auth();
+
+    helper.authEvents ={
+      clientCreated: () => 'client created',
+      clientDeleted: () => 'client deleted',
+    };
+
+    class EventHandler {
```
djmitche: This appears to be a fake version of [PulseTestReceiver](https://www.npmjs.com/package/taskcluster-lib-testing#pulsetestreceiver) which is already a testing utility and doesn't need to be faked :)

fiennyangeln:
@djmitche Sorry for the delay!  I have tried creating the fake publisher, based on main.js, then i try to slowly move the one in the else block to the outside. However, I cant move the webServer part (https://github.com/fiennyangeln/taskcluster-auth/blob/make-test-independent/test/helper.js#L129) which i think is because it called 'server' which needs 'api' which called this line (https://github.com/fiennyangeln/taskcluster-auth/blob/make-test-independent/src/main.js#L167) which needs pulse username and password.  What do you think about this?

fiennyangeln:
@djmitche some tests are failing because the webserver is not the real web server. How should I go about it?

src/v1.js
```diff
@@ -20,6 +20,7 @@ const roleToJson = (role, context) => _.defaults(
 /** API end-point for version v1/ */
 var api = new API({
   title:      'Authentication API',
+  name:       'auth',
```
jonasfj: Move to a separate PR, which I suspect we already have.

test/testserver.js
```diff
@@ -11,6 +11,7 @@ const PORT = 60321;
 let myapi = new API({
   title:        'Test API Server',
   description:  'API server for testing',
+  name:         'auth',
```
jonasfj: wrong, this is a different API, not that it matters... name could be `testservice`

src/v1.js
fiennyangeln: yes it is merged on other PR. will remove it.

test/testserver.js
fiennyangeln: yes it is merged on other PR. will remove it. :)

fiennyangeln:
@jonasfj 
1. do we need to create fake `helper.events` as it is called in several tests and now it is not created if there is no pulse credentials (https://github.com/fiennyangeln/taskcluster-auth/blob/make-test-independent/test/helper.js#L44 ) ? I tried `helper.events = new testing.PulseTestReceiver({fake:true}, mocha);` but it doesnt seem to work as expected (gives me connect ECONNREFUSED 127.0.0.1:5672 ) 
2. could you tell me more about the key routing module?
I will try to look at fake-amqp implementation :D thanks!!

jonasfj:
> helper.events = new testing.PulseTestReceiver({fake:true}, mocha);

Obviously, support for doing this would need to be added to `taskcluster-lib-testing`.

test/helper.js
```diff
@@ -58,6 +62,36 @@ class FakeRoles {
   }
 }
 
+class FakePublisher {
+  constructor() {
+    this.clients = [];
+  }
+
+  async clientCreated(clientId) {
```
djmitche: Note that all of these functions take an object containing the arguments, so `async clientCreated({clientId})`

test/helper.js
```diff
@@ -58,6 +62,36 @@ class FakeRoles {
   }
 }
 
+class FakePublisher {
+  constructor() {
+    this.clients = [];
```
djmitche: My approach here would be to set `this.calls = []` and then push something in each message-sending function, e.g.

```javascript
async clientUpdated({clientId}) {
  this.calls.push({method: 'clientUpdated', clientId});
}
```

Then later you can assert something about the contents of `helper.publisher.calls`.

test/helper.js
```diff
@@ -87,56 +121,58 @@ mocha.before(async () => {
...
   }
 
+  overwrites.publisher = new FakePublisher();
+
+  overwrites.resolver = helper.resolver =
+    await serverLoad('resolver', overwrites);
+
   if (!helper.hasPulseCredentials()) {
-    return;
+    webServer = null;
```
djmitche: You should be able to create the webServer whether there are pulse credentials or not.

fiennyangeln: But to create the server, we need api, then in the api it seems to take create a PulseConnection https://github.com/fiennyangeln/taskcluster-auth/blob/make-test-independent/src/main.js#L167 .. thats why I try to override the cfg.pulse value

fiennyangeln:
@jonasfj @djmitche I think the first one might be the best if we have the resources. What do you think?

djmitche:
The downside with the first option is that it doesn't work unless you have pulse credentials -- which makes it impossible to run the tests locally.  I like the third option, personally.

test/helper.js
djmitche: Oh, that's kind of gross.  How about this: let's move the pulse connection to a separate loader component (parallel to resolver, Client, Roles, etc.).  Then in the tests we can make a [fake version](https://github.com/taskcluster/taskcluster-client#fake-listening).  For the API tests, we don't need to actually send any messages to it -- just feeding the resolver a fake PulseConnection will be enough.

fiennyangeln:
@djmitche For the third option, should we use https://github.com/mochajs/mocha/wiki/Spies to know if the function gets called?

djmitche:
Yes, sinon and tools like it are a good way to track calls.  Usually I just do it manually (for example, I suggested `helper.publisher.calls` above), but that's just my habit.

test/helper.js
```diff
@@ -37,11 +40,12 @@ helper.hasPulseCredentials = function() {
...
   return cfg.app.hasOwnProperty('azureAccounts') && cfg.app.azureAccounts;
 };
-
 // Configure PulseTestReceiver
 if (cfg.pulse.password) {
   helper.events = new testing.PulseTestReceiver(cfg.pulse, mocha);
-};
+} else {
+  helper.events = new testing.PulseTestReceiver({fake:true}, mocha);
+}
```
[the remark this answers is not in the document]
fiennyangeln: @djmitche Oh. So since helper.events is removed, then the waitFor method by the helper.events will also be deleted and it does not matter since we also use fake pulse in this case and thus it wont take a while to return, is it?

test/helper.js
fiennyangeln: Noted!

test/helper.js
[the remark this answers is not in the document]
djmitche: correct!

fiennyangeln:
@djmitche does this change should let the test run when user does not have both azureAccount and pulse credentials? I see there is an in-memory implementation of the azure, so initially i think what i need to do is move the pulse connection to separate component as you suggested,

`  connection: {
    requires: ['cfg'],
    setup: async ({ cfg }) => {
      return new taskcluster.PulseConnection(cfg.pulse)
    },
  },` 
but then, to create the api, we need the azureAccount https://github.com/fiennyangeln/taskcluster-auth/blob/make-test-independent/src/main.js#L182

then I tried to overwrite the azureAccount using process.env like 

`process.env.AZURE_ACCOUNTS = {taskclusterdev: "test"}`  is it the correct way?

or should we overwrite the whole api?

djmitche:
The Azure stuff is already faked out in [helper.js](https://github.com/fiennyangeln/taskcluster-auth/blob/6ae650e838b61124bafd2b3b2d0d99867b6fb8fc/test/helper.js#L104-L122), so you shouldn't need to do anything with that.  Just setting up the fake connection should do the trick.

fiennyangeln:
@djmitche i have finished faking the connection. But I got this error "AssertionError [ERR_ASSERTION]: Context must have declared property: 'azureAccounts'"  which i think is because when we load api we need Azureaccount ( https://github.com/fiennyangeln/taskcluster-auth/blob/make-test-independent/src/main.js#L189 ) as it is specified in the context, thus it must be present.

djmitche:
Rebasing onto the latest master might help with that -- we've upgrade tc-lib-api.  if not, then passing an empty value for that in `helper.js` should help.

If not, including that value in your [user-config.yml](https://github.com/taskcluster/taskcluster-auth/blob/master/user-config-example.yml#L4), even with dummy information, should help.

fiennyangeln:
@djmitche I am able to make clienttest works now. Please have a look :) I will try to update the other tests.

test/client_test.js
```diff
@@ -177,29 +178,28 @@ suite('api (client)', function() {
       expires, description, scopes,
     });
 
-    await helper.events.waitFor('created');
+    assert.deepEqual(helper.publisher.calls, [{method: 'clientDeleted', clientId: CLIENT_ID},
+      {method: 'clientDeleted', clientId: CLIENT_ID}]);
   };
 
   test('auth.resetAccessToken', async () => {
-    await createTestClient();
```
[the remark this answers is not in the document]
fiennyangeln: Hi Dustin, sorry for the delay. It runs just fine in my laptop, I removed it because I get this error : 

`Error: client with same clientId already exists, possibly an issue with retry logic or idempotency
----
method:     createClient
errorCode:  RequestConflict
statusCode: 409
time:       2018-04-12T04:11:53.069Z
    at node_modules/taskcluster-client/lib/client.js:318:21
    at <anonymous>
    at process._tickDomainCallback (internal/process/next_tick.js:228:7)`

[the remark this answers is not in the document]
djmitche: That's pretty weird.  If you look at the top of the file, there's a `cleanup` function there which is invoked before and after each test script, and deletes all clients except `root`.  Perhaps you had that commented out when you were testing?

fiennyangeln:
@djmitche thanks for pointing out about the cleanup. I put it inside an else block which only run if we has pulse credentials. That's why it did not run. I changed the code, modify some lint issue, and remove the helper.authEvents part when it get called. Please review when u have time :D

test/helper.js
```diff
@@ -58,6 +54,51 @@ class FakeRoles {
   }
 }
 
+class FakePublisher {
```
jonasfj: For this to be a generic pattern that can be repeated in other services,  the `FakePublisher` class has to be derived from the `Exchanges` definition in `exchanges.js`.

This means tweaking the `pulse-publisher` package so that creation of such a class is supported. In effect we should have `const FakePublisher = require('../src/exchanges').createFakePublisher()`

jonasfj:
For the record, there is good work here, and we should land it :)

We do the generic solution in this PR, or in a follow up.

djmitche:
Yes, I'd like to use this as a model as we start to implement the new testing best-practices.  That might mean moving FakePublisher, or something like it, into taskcluster-lib-testing.

All that's left is a rebase, I think!

jonasfj:
> That might mean moving FakePublisher, or something like it, into taskcluster-lib-testing.

It could be built from references in `taskcluster-lib-testing`, but it's probably more sane to just build in `pulse-publisher`.

------
on topic, I have one slight problem with this...
 1. All tests have to be rewritten, arguably they'll be simpler than what we have with `WaitFor` (but we used to do some serious gymnastics in tc-queue and other projects, so that'll be fun) 
 2. You can write tests that wait for something to happen in a background job... 
 3. You are include to test for all messages published, instead of checking that a message was or was not published (this can be done by making different assertions).

I don't think any of these concerns should block landing this. But (2) might be considered when this is implemented in `pulse-publisher`, ideally soon :)

fiennyangeln:
@djmitche thanks for the comment. I tried rebasing it but it ended up failing the tests.. i am trying to get it back running:D

fiennyangeln:
@djmitche most of the tests case fail as i have insufficient scopes

Error: You do not have sufficient scopes. You are missing the following scopes:

{
  "AllOf": [
    "auth:create-client:nobody/sds:ad_asd/df-sAdSfchsdfsdfs",
    "scope1",
    "myapi:*"
  ]
}

djmitche:
Jonas and I talked about his concerns -- it's basically that this will probably need some refinement to apply to all tests in all repos, and I think that's fine.  It's a great start :)

I'm looking at the test failures now.

djmitche:
It looks like some of https://github.com/taskcluster/taskcluster-auth/commit/28176b01f44f922ae4685ca8d794ca3740077317#diff-5ab3980e63f908d4556c516d8c88bdc3 got lost in the rebase.  I'll push a fix.

fiennyangeln:
Thank you!! @djmitche

PR #144, opened 2018-03-21
upgrade to version 8.0.0 of taskcluster-lib-api

kritisingh1:
Shift to latest version of tc-lib-api

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "superagent": "^3.8.1",
     "superagent-hawk": "0.0.6",
     "taskcluster-client": "^3.1.0",
-    "taskcluster-lib-api": "^7.0.1",
+    "taskcluster-lib-api": "8.0.0",
     "taskcluster-lib-app": "^2.0.3",
     "taskcluster-lib-docs": "^4.0.0",
     "taskcluster-lib-iterate": "^1.0.2",
--- a/src/v1.js
+++ b/src/v1.js
@@ -110,6 +110,7 @@ var api = new API({
     // A tc-lib-monitor for use beyond the lib-api level
     'monitor',
 
+    // The webhooktunnel config (with properties `secret` and `proxyUrl`)
     'webhooktunnel',
   ],
 });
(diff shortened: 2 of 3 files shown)


djmitche:
Tests fail with
```
      AssertionError [ERR_ASSERTION]: Context has unexpected property: webhooktunnel
```
(/cc @ckousik)

kritisingh1:
The local tests passed after the change so I didn't look into this. Probably this is a problem which is common to all such upgrades. I will fix it asap. :)

kritisingh1:
@djmitche I have updated the pull request. However, since I can't see the results of the travis tests, they are likely to fail even now. The fix is however, declaring them as an array as the value of context where the API is declared. Please review. Thanks :)

djmitche:
This looks good.  I'll deploy tomorrow, if I have time.  In the interim, @jonasfj maybe you have a sentence about why this context value is required here?

jonasfj:
The `context` array of string properties is basically a way to force people to think about the `context` object they pass when they instantiate the API... in the `.setup({..., context: ...})`.

By forcing you to write a string I'm sure that you didn't just inject a lot of objects with no though as to what they do or how they are named... it also encourages writing a comment... and services as documentation, when hacking on API implementation it's easy to look at the `context` declared to see what properties are available on `this`.

In short, this could all be dynamically typed, but by forcing you to be a little bit more explicit we catch a lot of bugs. Often we forget to set a property declared to be in the `context` object, and the `.setup` method will complain... which is pretty powerful... Otherwise, such bugs can be hard to catch and mismatch or typos only show up at runtime.

djmitche:
Sorry, yes, that's why we're adding this check.  The question was why this *specific* key, `webhooktunnel`, was required.

jonasfj:
Probably it only became required that these keys are specified now... and `webhooktunnel` was added before that... I'm sure it's a config section with server url and shared secret for webhooktunnel, so that auth can issue jwt for workers.

djmitche:
Yes, that's the change @kritisingh1 is making now :)

djmitche:
I pushed a change to add a comment about that property.  I'll get this merged soon!

djmitche:
..and deployed, thanks!

PR #146, opened 2018-03-27
added support for static clients

jonasfj:


src/data.js
```diff
@@ -105,39 +105,88 @@ Client.prototype.json = function(resolver) {
...
+
+  // Create new clients
+  await Promise.all(newClients.map(target => {
+    return this.create({
+      clientId:     target.clientId,
+      description:  target.description + descriptionSuffix,
+      accessToken:  target.accessToken,
+      expires:      taskcluster.fromNow('1000 year'),
+      scopes:       target.scopes,
+      disabled:     0,
```
djmitche: `false`

jonasfj: Looks like it's declared `Entity.types.Number`, I bet this is from before we had `Entity.types.Boolean`.

I agree we should change this, but not in this PR.

jonasfj:
I added some tests for the sync process..

src/data.js
djmitche: Agreed

src/data.js
```diff
@@ -105,39 +105,88 @@ Client.prototype.json = function(resolver) {
...
+    assert(typeof client.description === 'string', 'expected description to be a string');
+    assert(client.scopes instanceof Array, 'expected scopes to be an array of strings');
+    assert(client.clientId.startsWith('static/'), 'static clients must have clientId = "static/..."');
+    assert(client.scopes.every(s => typeof s === 'string'), 'scopes must be strings');
+  }
+
+  // description suffix to use for all static clients
+  const descriptionSuffix = [
+    '',
+    '---',
```
djmitche: This should have newlines before and after, otherwise it shows up as a markdown heading

config.yml
```diff
@@ -20,9 +20,20 @@ defaults:
...
     statsComponent:           auth
 
-    # root accessToken, if defined will cause root client to be
-    # automatically created when application is started
-    rootAccessToken:          !env ROOT_ACCESS_TOKEN
+    # Static clients, these clients are built-in from start, and can't be
+    # modified or deleted at runtime. Useful for granting clientIds to other
+    # services and boostrapping with a root client.
+    # Each entry must have: clientId, accessToken, description, scopes
+    # where clientId MUST start with static, by convention these namespaced:
```
djmitche: ..start with static/

jonasfj:
Okay, I'll deploy tomorrow...

PR #148, opened 2018-03-28
added paginateListClients API and tests

tuhina2020:
BUG : https://bugzilla.mozilla.org/show_bug.cgi?id=1436212

### DIFF
--- a/src/v1.js
+++ b/src/v1.js
@@ -135,29 +135,37 @@ api.declare({
   description: [
     'Get a list of all clients.  With `prefix`, only clients for which',
     'it is a prefix of the clientId are returned.',
+    '',
+    'By default this end-point will try to return up to 1000 clients in one',
+    'request. But it **may return less, even none**.',
+    'It may also return a `continuationToken` even though there are no more',
+    'results. However, you can only be sure to have seen all results if you',
+    'keep calling `listClients` with the last `continuationToken` until you',
+    'get a result without a `continuationToken`.',
   ].join('\n'),
 }, async function(req, res) {
   let prefix = req.query.prefix;
   let continuationToken  = req.query.continuationToken || undefined;
-  let limit         = parseInt(req.query.limit || 10, 10);
+  let limit         = parseInt(req.query.limit || 1000, 10);
   let Client = this.Client;
   let resolver = this.resolver;
 
   let response = {clients: []};
 
-  while (true) {
-    let data = await Client.scan({}, {limit, continuation: continuationToken});
-    data.entries.forEach(client => {
-      if (!prefix || client.clientId.startsWith(prefix)) {
-        response.clients.push(client.json(resolver));
-      }
-    });
-    continuationToken = data.continuation || undefined;
-    if (!_.isEmpty(response.clients) || !continuationToken) {
-      break;
-    }
+  let opts = {limit};
+  if (req.query.continuationToken) {
+    opts.continuation = req.query.continuationToken;
   }
-  response.continuationToken = continuationToken;
+  let data = await Client.scan({}, opts);
+  data.entries.forEach(client => {
+    if (!prefix || client.clientId.startsWith(prefix)) {
+      response.clients.push(client.json(resolver));
+    }
+  });
+  if (data.continuation) {
+    response.continuationToken = data.continuation;
+  }
+
   res.reply(response);
 });
 
--- a/src/v1.js
+++ b/src/v1.js
@@ -138,35 +138,35 @@ api.declare({
   ].join('\n'),
 }, async function(req, res) {
   let prefix = req.query.prefix;
-  let continuation  = req.query.continuationToken || undefined;
+  let continuationToken  = req.query.continuationToken || undefined;
   let limit         = parseInt(req.query.limit || 10, 10);
   let Client = this.Client;
   let resolver = this.resolver;
 
-  const method = async function(limit, continuation) {
-    let response = {clients: [], continuation};
-    let data = await Client.scan({}, {limit, continuation});
+  const method = async function(limit, continuationToken) {
+    let response = {clients: [], continuationToken};
+    let data = await Client.scan({}, {limit, continuation: continuationToken});
     data.entries.forEach(client => {
       if (!prefix || client.clientId.startsWith(prefix)) {
         response.clients.push(client.json(resolver));
       }
     });
-    response.continuation = data.continuation;
+    response.continuationToken = data.continuation;
     return response;
   };
 
-  const retryMethod = async function(limit, continuation, retries) {
+  const retryMethod = async function(limit, continuationToken, retries) {
     let i = 1;
-    let response = await method(limit, continuation);
+    let response = await method(limit, continuationToken);
     while (i <= retries && _.isEmpty(response.clients)) {
-      continuation = response.continuation;
-      response = await method(limit, continuation);
+      continuationToken = response.continuationToken;
+      response = await method(limit, continuationToken);
       i++;
     }
     return response;
   };
 
-  let response = await retryMethod(limit, continuation, 1);
+  let response = await retryMethod(limit, continuationToken, 1);
   res.reply(response);
 });
 
--- a/src/v1.js
+++ b/src/v1.js
@@ -151,6 +151,42 @@ api.declare({
   res.reply(clients);
 });
 
+/** Paginate List clients */
+api.declare({
+  method:     'get',
+  route:      '/clients/',
+  query: {
+    prefix:  /^[A-Za-z0-9!@/:.+|_-]+$/, // should match clientId above
+    continuationToken: /./,
+    limit: /^[0-9]+$/,
+  },
+  name:       'paginateListClients',
+  input:      undefined,
+  output:     'list-clients-response.json#',
+  stability:  'stable',
+  title:      'List Clients',
+  description: [
+    'Get a list of all clients.  With `prefix`, only clients for which',
+    'it is a prefix of the clientId are returned.',
+  ].join('\n'),
+}, async function(req, res) {
+  let prefix = req.query.prefix;
+  let continuation  = req.query.continuationToken || null;
+  let limit         = parseInt(req.query.limit || 1000, 10);
+
+  // Load all clients
+  // TODO: as we acquire more clients, perform the prefix filtering in Azure
+  let data = await this.Client.scan({}, { limit, continuation });
+  let response = { clients: [], continuation : data.continuation };
+  data.entries.forEach(client => {
+    if (!prefix || client.clientId.startsWith(prefix)) {
+      response.clients.push(client.json(this.resolver));
+    }
+  });
+
+  res.reply(response);
+});
+
 /** Get client */
 api.declare({
   method:     'get',
--- a/src/v1.js
+++ b/src/v1.js
@@ -124,6 +124,8 @@ api.declare({
   route:      '/clients/',
   query: {
     prefix:  /^[A-Za-z0-9!@/:.+|_-]+$/, // should match clientId above
+    continuationToken: /./,
+    limit: /^[0-9]+$/,
   },
   name:       'listClients',
   input:      undefined,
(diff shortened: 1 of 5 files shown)


src/v1.js
```diff
@@ -148,6 +148,42 @@ api.declare({
   res.reply(clients);
 });
 
+/** Paginate List clients */
```
djmitche: This should be a modification of the existing method, rather than a new method ([ideally](https://bugzilla.mozilla.org/show_bug.cgi?id=1449697), `api.declare` would throw an error if you declare the same route+method twice).

src/v1.js
```diff
@@ -148,6 +148,42 @@ api.declare({
...
+
+  // Load all clients
+  // TODO: as we acquire more clients, perform the prefix filtering in Azure
+  let data = await this.Client.scan({}, { limit, continuation });
+  let response = { clients: [], continuation : data.continuation };
+  data.entries.forEach(client => {
+    if (!prefix || client.clientId.startsWith(prefix)) {
+      response.clients.push(client.json(this.resolver));
+    }
+  });
```
djmitche: This looks great :)

tuhina2020:
Modified the existing `listClients` API, instead of creating a new one. https://github.com/taskcluster/taskcluster-auth/pull/148/files#diff-dec5765e80c6f195a1d17a86ec5835b1R121

djmitche:
The credentials I emailed to you should help -- it should be using `jungle` instead of `taskclusterdev`.

test/client_test.js
```diff
@@ -115,14 +115,14 @@ suite('api (client)', function() {
       c => c.clientId.substr(CLIENT_ID.length)).sort();
 
     // get all clients
-    assume(gotSuffixes(await helper.auth.listClients())).to.deeply.equal(suffixes);
```
djmitche: Oh dear, I missed this in the last round of review.  This is bad -- it means anything that expects to get a plain list of clients back will fail.

@jonasfj what do you think -- how should we solve this?

src/v1.js
```diff
@@ -133,19 +135,30 @@ api.declare({
...
+  let limit         = parseInt(req.query.limit || 1000, 10);
 
-  // Load all clients
-  // TODO: as we acquire more clients, perform the prefix filtering in Azure
-  let clients = [];
-  await this.Client.scan({}, {
-    handler: client => {
-      if (!prefix || client.clientId.startsWith(prefix)) {
-        clients.push(client.json(this.resolver));
+  const retryMethod = async function(limit, continuation, retries) {
```
jonasfj: Why this?

test/client_test.js
jonasfj: If we fix tc-tools, and tc-admin + tc-cli, I think we're safe... But we should probably fix those to handle both cases first..

That's just a suggestion, it might also be acceptable to break... However, given that the API is declared stable... Perhaps we should make some noise about this change on a mailing list... Ask if anyone is worried...

Maybe look through logs to see how much it's used...

djmitche:
Yes, the patch still needs some work -- just looking for feedback on the compatibility issue.

I'm worried that we wouldn't catch all cases with the breakage.  And especially in a redeployable implementation, we can't use that technique.

Another option is to create a new URL for the new, paginated version.  The existing API endpoint is at `/v1/clients` so we could put the paginated endpoint at `/v1/clients/paginated` or something like that.  Then users still hitting the old endpoint would get the expected response.

We can do a little better here by changing the name of the old API method to `listAllClients` and naming the new endpoint `listClients`.  Then users of libraries like taskcluster-client, taskcluster-cli, etc. will automatically use the new endpoint (since they are calling endpoints by name) and in a year or two we can remove the old endpoint.

The only downside is that `/v1/clients/paginated` isn't a very RESTful name.  Is that OK?

jonasfj:
This would break `taskcluster-login` which is security sensitive... so we should definitely keep this in mind.

tuhina2020:
@jonasfj Changed the list-clients-response schema. @djmitche let me know about hte changes the patch needs.

jonasfj:
I think we should patch all consumers of the API before trying to land this change.
Such that consumers can handle new and old API, but testing if the result is an object or an array.

I think the consumers are:
 * taskcluster-login
 * taskcluster-tools (client manager, scope inspector, cli login flow)

Note: this is also a good lesson for the future, to always return an object as it makes introducing new properties easier :)

djmitche:
OK, so to put it another way, we'll make a change that breaks compatibility, first fixing all of the uses of the method that we know of to accept either the old (just a list) or new (object with a `clients` property) format.  It looks like there are only the two places @jonasfj mentioned that need to be updated.

@tuhina2020 does that make sense?  Do you want to get started by making PR's to login and tools?

tuhina2020:
@djmitche Yes sure. I will fix the calls to listClients in the above repositories first.

src/v1.js
```diff
@@ -133,19 +135,36 @@ api.declare({
...
+      continuationToken = response.continuationToken;
+      response = await method(limit, continuationToken);
+      i++;
+    }
+    return response;
+  };
 
-  res.reply(clients);
+  let response = await retryMethod(limit, continuationToken, 1);
+  res.reply(response);
```
djmitche: I think this could be implemented more simply with a `while` loop, instead of two nested functions.  Something like

```
while (true) {
  scan with continuationToken / limit
  if results
    return them
  else
    update continuationToken
}
```

Do you want to try refactoring along those lines?  The nice thing about having tests written is, you'll know immediately if your refactor works or not :)

tuhina2020: @djmitche refactored the code as you suggested.

src/v1.js
```diff
@@ -136,19 +138,27 @@ api.declare({
...
+  let limit         = parseInt(req.query.limit || 10, 10);
+  let Client = this.Client;
+  let resolver = this.resolver;
 
-  // Load all clients
-  // TODO: as we acquire more clients, perform the prefix filtering in Azure
-  let clients = [];
-  await this.Client.scan({}, {
-    handler: client => {
+  let response = {clients: [], continuationToken};
```
djmitche: We'll want to omit continuationToken here.  Otherwise the caller will get back the continuationToken they just sent, and loop forever!

src/v1.js
```diff
@@ -136,19 +138,27 @@ api.declare({
...
+
+  while (true) {
+    let data = await Client.scan({}, {limit, continuation: response.continuationToken});
```
djmitche: Let's use `continuationToken` here (the variable, rather than the property of `response`).

src/v1.js
```diff
@@ -136,19 +138,27 @@ api.declare({
...
+    data.entries.forEach(client => {
       if (!prefix || client.clientId.startsWith(prefix)) {
-        clients.push(client.json(this.resolver));
+        response.clients.push(client.json(resolver));
       }
-    },
-  });
-
-  res.reply(clients);
+    });
```
djmitche: And here, before checking whether it's empty, set `continuationToken` to `data.continuationToken`.  Then if we loop again, we'll use the new continuationToken.

If `continuationToken` is not set at this point, we need to break from the loop and return an empty list to the user.

src/v1.js
```diff
@@ -136,19 +138,27 @@ api.declare({
...
+    if (!_.isEmpty(response.clients)) {
+      break;
+    } else {
+      response.continuationToken = data.continuation;
+    }
+  }
+  res.reply(response);
```
djmitche: Set `response.continuationToken = continuationToken` here, just before `res.reply`.

src/v1.js
```diff
@@ -136,19 +138,27 @@ api.declare({
...
       }
-    },
-  });
-
-  res.reply(clients);
+    });
+    if (!_.isEmpty(response.clients)) {
+      break;
+    } else {
+      response.continuationToken = data.continuation;
```
djmitche: You won't need this `else` anymore.

test/client_test.js
```diff
@@ -110,20 +110,21 @@ suite('api (client)', function() {
...
-    assume(gotSuffixes(await helper.auth.listClients({prefix: CLIENT_ID + '/bb/'})))
-      .to.deeply.equal(['/bb/1', '/bb/2']);
-    assume(gotSuffixes(await helper.auth.listClients({prefix: CLIENT_ID + '/c'})))
-      .to.deeply.equal([]);
+    // assume(gotSuffixes(await helper.auth.listClients({prefix: CLIENT_ID + '/bb'})))
+    //   .to.deeply.equal(['/bb', '/bb/1', '/bb/2']);
+    // assume(gotSuffixes(await helper.auth.listClients({prefix: CLIENT_ID + '/bb/'})))
+    //   .to.deeply.equal(['/bb/1', '/bb/2']);
+    // assume(gotSuffixes(await helper.auth.listClients({prefix: CLIENT_ID + '/c'})))
+    //   .to.deeply.equal([]);
```
djmitche: do these still pass?

test/helper.js
```diff
@@ -95,7 +95,7 @@ mocha.before(async () => {
 
     webServer = await serverLoad('server', overwrites);
     webServer.setTimeout(3500); // >3s because Azure can be sloooow
-
+
```
djmitche: (extra whitespace)

test/client_test.js
djmitche: Sorry, I forgot to add -- please make some tests that check that continuationToken and limit work, along with filtering.  You can use `limit: 1` to make each batch of results come back with at most one client in it.

src/v1.js
tuhina2020: Implemented this refactor.

test/client_test.js
```diff
@@ -110,20 +110,34 @@ suite('api (client)', function() {
...
-    assume(gotSuffixes(await helper.auth.listClients())).to.deeply.equal(suffixes);
+    let clients = await helper.auth.listClients();
+    assume(gotSuffixes(clients)).to.deeply.equal(suffixes);
 
     // prefix filtering
     assume(gotSuffixes(await helper.auth.listClients({prefix: CLIENT_ID + '/bb'})))
-      .to.deeply.equal(['/bb', '/bb/1', '/bb/2']);
+      .to.deeply.equal(['/bb', '/bb/1', '/bb/2', '/bb/3', '/bb/4', '/bb/5']);
     assume(gotSuffixes(await helper.auth.listClients({prefix: CLIENT_ID + '/bb/'})))
       .to.deeply.equal(['/bb/1', '/bb/2']);
```
djmitche: Hm, I get 
```
     Unknown assertation failure occured, assumed `[ '/bb/1', '/bb/2', '/bb/3', '/bb/4', '/bb/5' ]` to deeply equal `[ '/bb/1', '/bb/2' ]`
```
on this line.  Which makes sense, right?

test/client_test.js
```diff
@@ -110,20 +110,34 @@ suite('api (client)', function() {
...
     assume(gotSuffixes(await helper.auth.listClients({prefix: CLIENT_ID + '/c'})))
       .to.deeply.equal([]);
+    assume(gotSuffixes(await helper.auth.listClients({
+      prefix: CLIENT_ID + '/bb',
+      limit: 1,
+    }))).to.deeply.equal(['/bb']);
```
djmitche: Azure doesn't actually guarantee the order that results are returned.  So this could be /bb, or it could be /bb/3.

test/client_test.js
```diff
@@ -110,20 +110,34 @@ suite('api (client)', function() {
...
+    assume(gotSuffixes(await helper.auth.listClients({
+      prefix: CLIENT_ID + '/bb',
+      limit: 3,
+    }))).to.deeply.equal(['/bb', '/bb/1', '/bb/2']);
```
djmitche: Same issue here.

src/v1.js
```diff
@@ -136,19 +138,27 @@ api.declare({
...
+  let resolver = this.resolver;
 
-  // Load all clients
-  // TODO: as we acquire more clients, perform the prefix filtering in Azure
-  let clients = [];
-  await this.Client.scan({}, {
-    handler: client => {
+  let response = {clients: []};
+
+  while (true) {
```
jonasfj: The `while(true)` thing here isn't necesary... I think it's okay to return `{clients: [], continuationToken: '...'}`.

Mostly, I think it's better than risk the a timeout, which is what `continuationToken` is intended to prevent.

djmitche: Hm, I thought the opposite, but you make a good point about timeouts -- if we were to have a filter that matched only one client, and a LOT of clients, then it might take more than 30s to find that one client.  We have a limit of 30s for each HTTP transaction.

@tuhina2020 rather than making you reverse that change, I'll push a commit that does it.  Sorry!

djmitche:
OK, that looks better and passes in my testing.  Note that I also increased the default for limit from 10 to 1000.  Leaving that at 10 would have caused existing (non-paginated) listClients calls only only return 10 results!

PR #150, opened 2018-04-13
Bug 1452000 - update to new lib-monitoring and supply project, enable…

djmitche:
… from config


djmitche:
Oops, hang on..

djmitche:
ok, ready for review now :)

djmitche:
I already updated the config, but maybe I could avoid that necessity with `|| 'taskcluster-auth'`.  That probably wouldn't hurt.

djmitche:
This is going to upgrade raven as well, so I'd like to hold off.  I'll try it on a less critical bug first.

PR #151, opened 2018-04-16
Cleaned up schemas to improve generated go types in taskcluster-client-go

petemoore:
This names types to avoid that we generate types `Var`, `Var1`, `Var2`, `Var3` when using non-nested structs in taskcluster-client-go.

Names that are too long are discarded, so this shortens two names that were too long, and also gives names to `oneOf` subtypes.

### DIFF
--- a/schemas/authenticate-hawk-request.yml
+++ b/schemas/authenticate-hawk-request.yml
@@ -43,9 +43,10 @@ properties:
       **Note,** order of querystring elements is important.
   host:
     type:                 string
-    anyOf:
-      - format:           hostname
-      - format:           ipv4
+    # IPv4 addresses conform to `hostname` format according to jsonschema, so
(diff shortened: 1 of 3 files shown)


schemas/authenticate-hawk-request.yml
```diff
@@ -42,13 +42,19 @@ properties:
       string that follows the HTTP method.
       **Note,** order of querystring elements is important.
   host:
-    type:                 string
```
jonasfj: How does removing this help?

Once the type generation code sees `type: string` it could ignore all the other properties... as all of the `oneOf/anyOf` options are unimportant... If we have `type: string` we make it a string (unless we have `type: string, format: date-time`)

But `type: string, oneOf: ...` should just be a string... wouldn't that be better to fix in the codegen part?

PR #152, opened 2018-04-26
Update hoek for CVE-2018-3728

imbstack:


imbstack:
I don't think it ever returns one. Happy to simplify it or just get it in as-is.

PR #154, opened 2018-04-28
Bug 1457608: Make signatures work in heroku and kubernetes

imbstack:
This hacks the things to work in both places. I have not put much thought into any of the names or whatever of this, but at least it works? I think this will fail tests and stuff still. Just getting it out there for our monday discussion.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -95,6 +95,9 @@ defaults:
     # Trust a forwarding proxy
     trustProxy:               !env:bool TRUST_PROXY
 
+    # TODO: pick a better name for this
+    proxyName:                !env PROXY_NAME
+
   azure:
     # Azure table storage account name
     accountName:              !env AZURE_ACCOUNT_NAME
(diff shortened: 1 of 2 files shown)


config.yml
```diff
@@ -95,6 +95,9 @@ defaults:
     # Trust a forwarding proxy
     trustProxy:               !env:bool TRUST_PROXY
 
+    # TODO: pick a better name for this
+    proxyName:                !env PROXY_NAME
```
jonasfj: May I suggest `pathPrefix` or something else? maybe `vhost` (probably bad).

If you want to leave this with a deliberately ugly name, please add a comment explaining what it is..


----
I think this often called `virtualPath`

imbstack: I think if we do this right, we can actually _simplify_ this whole process and perhaps even make the `/v1` bit useful all at one fell swoop.

I also notice that at least in auth, `publicUrl` does pretty much nothing other than be in schema IDs. I'm thinking that maybe we can get rid of it and just do relative paths for that too.

imbstack:
We've fixed this for real now!

PR #156, opened 2018-05-10
Increase length of project names to allow for standardizing on projectName

imbstack:
Currently we limit this to 22 chars and I think we only do that for "consistency". We end up having lots of weird project names to fit into this limit. May as well raise it?

**NOTE: We must set TASKCLUSTER_ROOT_URL env var to `https://taskcluster.net` before we deploy this**

### DIFF
--- a/src/v1.js
+++ b/src/v1.js
@@ -81,7 +81,7 @@ var api = new API({
     // doesn't work well with HTTPS and virtual-style hosting.
     // Hence, we shouldn't encourage people to use them
     // Project for sentry (and other per project resources)
-    project:    /^[a-zA-Z0-9_-]{1,22}$/,
+    project:    /^[a-zA-Z0-9_-]{1,64}$/,
   },
   context: [
     // Instances of data tables


config.yml
```diff
@@ -128,6 +130,8 @@ production:
     env:                      'production'
     development:              false
 test:
+  taskcluster:
+    rootUrl:                  http://example.com
```
djmitche: shouldn't this be the testRootUrl?

schemas/statsum-token-response.yml
```diff
@@ -26,9 +26,16 @@ properties:
     title:          BaseUrl
     description: |
       Base URL for the statsum server this project is allocated on.
+  rootUrl:
+    type:           string
+    format:         uri
+    title:          rootUrl
+    description: |
+      Root URL for a Taskcluster cluster.
```
djmitche: Two questions here:

 1. Will statsum be treated like all of the other services, even though it's in Go and not JS (in particular, will it be at `:rootUrl/api/v1/project/:project`?)
 1. If so, do we want to allow for that to run on a different rootUrl than the rest of the cluster?

My thinking is that if the answer to (1) is "yes", then we should define this response as always applying to the "statsum' service in the same instance of Taskcluster as the auth service -- so, no need to return a rootUrl or a baseUrl.  If the answer to (1) is "no", then perhaps we can rename baseUrl to something more specific like statsumUrl and define this response as returning a token valid at `:statsumUrl/v1/project/:project`.

I actually don't care either way what the answer to (1) is.  "Yes" would not be too difficult to implement, and would provide a nice starting place if we ever decide to start building more services in Golang.

src/v1.js
```diff
@@ -81,7 +81,7 @@ var api = new API({
     // doesn't work well with HTTPS and virtual-style hosting.
     // Hence, we shouldn't encourage people to use them
     // Project for sentry (and other per project resources)
-    project:    /^[a-zA-Z0-9_-]{1,22}$/,
```
djmitche: :+1:

schemas/statsum-token-response.yml
jonasfj: (1) is probably "no".

Reasoning being that we won't have load balancer in front of statsum (we can), but we wont have multiple instance behind the load balancer. If we need to load balance we would use a hash function to assign each `project` to a statsum server. If those statsum servers forward metrics to the same destination, that's no problem as long as they don't have the same project names.

If wanted statsum to be a TC service it should be accepting TC creds, if we want to do that, then it's a different service. In this case we should think of statsum the way we think of sentry.

PR #161, opened 2018-08-23
Bug 1440965 - enforce scopes don't end in **

djmitche:
@petemoore can you double-check that the golang regexp engine will handle the negative lookbehind here?

### DIFF
--- a/docs/scopes.md
+++ b/docs/scopes.md
@@ -33,3 +33,15 @@ wildcard, matching any suffix. So `queue:create-task:test-provisioner/*`
 satisfies `queue:create-task:test-provisioner/worker3`. The reverse is not
 true. The wildcard only works at the end of a string: no more advanced
 pattern-matching functionality is available.
+
+### Double-Stars
+
+The scope `foo:**` satisfies, on its face, any scope beginning with `foo:*`, including `foo:*123` and `foo:*` but not `foo:abc`.
+If scope `foo:**` is used to create a temporary credential with `foo:*`, the operation will succeed (`foo:**` satisfies `foo:*`).
+But the resulting temporary credential has `foo:*`, which *does* satisfy `foo:abc`, unlike the original credential.
+In fact, this can occur anywhere a scope is delegated, including temporary credentials, authorizedScopes, task creation, client creation, and role creation.
+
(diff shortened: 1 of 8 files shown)


petemoore:
> @petemoore can you double-check that the golang regexp engine will handle the negative lookbehind here?

For generated go types, there is no enforcement of pattern matching, the patterns are used purely for doc comment generation. In the case of arrays, we currently generate the docs based on the containing schema (parent of `items`), rather than the schema under `items` - i.e. at the moment the generated docs neglect to include the pattern in the generated comment. See [here](https://github.com/taskcluster/taskcluster-client-go/blob/140f11ad305d8b9c2f2fdbd5dc81de1009a28483/tcauth/types.go#L230-L234) as an example.

Previously we generated a type for the `items` subschema which had the advantage you could pin the documentation to a custom type, versus doing the simple thing of just having an array of a primitive type, such as `[]string`. For example, we would have previously created `type Scope string`, and its docs would have contained the information in the json subschema for items (such as the regular expression it should conform to), but if I remember correctly, jonasfj preferred that optimised out this step so that you would have `[]string` rather than `[]Scope`.

We could revisit that decision, but suffice to say, this only really affects documentation rather than functional behaviour of generated code.

Another place we process scopes is in the [scopes package](https://godoc.org/github.com/taskcluster/taskcluster-base-go/scopes). This library provides methods for scope expansion and scope satisfaction, so it probably doesn't need to be touched for this.

The other place that springs to mind where we do handle scopes is when generating temporary clients, which is done [here](https://github.com/taskcluster/taskcluster-client-go/blob/621310e14994f5040833a1c4f044984725f7e797/creds.go#L78-L146). We could potentially put some restrictions here. I'm in two minds about this, in that I don't want to build too much logic/magic into the client libraries, and prefer the services to manage scope validation where possible, keeping the clients as dumb as possible so that they are as robust as possible in the face of change. But we can discuss!

I can't think of any other code that may be impacted by this, at the moment.

djmitche:
I mentioned in the bug, I don't want to actually change the scope-satisfaction algorithms, since as you say their simplicity is more important than that they reject this edge case.  So I think all I care about is the docs :)

PR #163, opened 2018-09-04
Add sourceIp to auditlog log line

ajvb:
Accept `sourceIp` in `createSignatureValidator`'s `options` object and log that parameter within the audit log.

Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1488545
Related: https://github.com/taskcluster/taskcluster-lib-api/pull/116

### DIFF
--- a/src/signaturevalidator.js
+++ b/src/signaturevalidator.js
@@ -181,7 +181,7 @@ const limitClientWithExt = function(credentialName, issuingClientId, accessToken
  * }
  *
  * The function returned takes an object:
- *     {method, resource, host, port, authorization}
+ *     {method, resource, host, port, authorization, sourceIp}
  * And returns promise for an object on one of the forms:
  *     {status: 'auth-failed', message},
  *     {status: 'auth-success', clientId, scheme, scopes}, or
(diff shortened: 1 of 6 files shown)


src/signaturevalidator.js
```diff
@@ -292,6 +292,7 @@ const createSignatureValidator = function(options) {
           port: req.port,
           resource: req.resource,
           method: req.method.toUpperCase(),
+          sourceIp: req.sourceIp,
```
djmitche: For consistency, this should probably have `|| '0.0.0.0'` so that we always get a valid IP in this slot (but a very distinctive one for cases where we do not know).

ajvb:
@djmitche As you said, I seem to be getting test errors due to a problem with the schema after adding `sourceIp` to https://github.com/taskcluster/taskcluster-auth/blob/master/schemas/v1/authenticate-hawk-request.yml

```
Error: schema is invalid: data.properties['scopes'].items.pattern should match format "regex", data.properties['scopes'].items should be array, data.properties['scopes'].items should match some schema in anyOf
```

I'm having trouble figuring out what I need to fix/change to solve this, would you mind pointing me towards what is causing this failure?

djmitche:
Can you push the changes causing that issue?

ajvb:
@djmitche Actually, it seems like it is a problem with me running tests locally. I'm getting the same errors on `master`.

ajvb:
@djmitche Ok, I resolved the issue locally. I was using node 8.9, and after switching to using the latest node 8.x all of the tests passed.

PR #164, opened 2018-09-06
Bug 1481944 - upgrade hawk

djmitche:
This sort of freaks me out, but the tests still pass, at least.


djmitche:
I like how you think :)

PR #165, opened 2018-09-25
Update lib-api for audit log ip addresses

imbstack:


djmitche:
Huh, what's up with the tests failing, though?

PR #166, opened 2018-09-27
Fix azure table listing

imbstack:
Turns out pagination never worked here and we never noticed because we never paged. Recently enough tables got added to the testing account that the asserted table was on the second page!

### DIFF
--- a/src/azure.js
+++ b/src/azure.js
@@ -30,7 +30,7 @@ builder.declare({
   route:      '/azure/:account/tables',
   name:       'azureTables',
   query: {
-    continuationToken: /^[A-Za-z][A-Za-z0-9]{2,62}$/,
+    continuationToken: /^.*$/,
   },
   input:      undefined,
   output:     'azure-table-list-response.yml',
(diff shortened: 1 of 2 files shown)


src/azure.js
```diff
@@ -44,14 +44,12 @@ builder.declare({
   let account = req.params.account;
   let continuationToken  = req.query.continuationToken || null;
 
-  await req.authorize({account});
-
```
djmitche: This is no longer necessary with scopeExpressions, right?

src/azure.js
```diff
@@ -44,14 +44,12 @@ builder.declare({
...
   let table = new azure.Table({
     accountId:  account,
     accessKey:  this.azureAccounts[account],
   });
 
-  let result = await table.queryTables({continuationToken});
```
djmitche: OOPS!

src/azure.js
imbstack: Yeah. It should be fine. We'll get alerts if I'm wrong as well!

PR #173, opened 2018-10-18
Bug 1488789 - use tc-lib-pulse

djmitche:
For real this time!

 * Use configured pulse credentials, so no dependency on another service
 * Use ephemeral consumers for the "notification" pulse messages
 * Use a regular producer to send those notifications

What am I missing?

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,6 @@
     "lodash": "^4.17.4",
     "morgan-debug": "^2.0.0",
     "promise": "^8.0.1",
-    "pulse-publisher": "^10.0.2",
     "quick-lru": "^1.1.0",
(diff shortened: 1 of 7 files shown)


config.yml
```diff
@@ -95,6 +95,9 @@ defaults:
 
 
   pulse:
+    # This service, tc-auth, cannot use tc-pulse, because tc-pulse depends on
+    # tc-auth.  So this must be a statically configured pulse user with access
+    # to namepsace 'taskcluster-auth'.
```
imbstack: I guess we don't even need to mention tc-pulse now?

djmitche: Oh, good call..

djmitche:
It stuck!!

PR #174, opened 2018-10-18
Bug 1499701 - upgrade fast-azure-storage

djmitche:


djmitche:
whoops, I need to upgrade and release azure-entities first

PR #177, opened 2018-10-30
add whitespace

djmitche:


djmitche:
That is weird -- there must be some missed race condition in there.

djmitche:
Haha, I was going to file a bug for the flaky test, but now there's no link to it!

PR #179, opened 2018-10-31
Update taskcluster-lib-validate for better errors

djmitche:


djmitche:
deployed (along with the last few :heavy_check_mark: for auth)

PR #181, opened 2018-11-20
Bug 1505405 - upgrade tc-lib-pulse

djmitche:
This only affects the output references format (for RFC 128)


djmitche:
Bug 1509154 for that flaky test...

PR #189, opened 2018-12-13
Add listRoleIds() to paginate roles 

OjaswinM:
Add a new function that returns roleIds based on limit and continuationToken.

### DIFF
--- a/schemas/v1/list-role-ids-response.yml
+++ b/schemas/v1/list-role-ids-response.yml
@@ -1,0 +1,21 @@
+$schema:  http://json-schema.org/draft-06/schema#
+title:                      "Get Role Response"
+description: |
+  Get all details about a role
+type: object
+properties:
+  roleIds:
+    description: |
+      roleIds correspnding to the list of roles requested
+    type:                 array
+    items:              {type: string}
+  continuationToken:
+    title:                "Continuation Token"
+    type:                 string
+    description: |
+      A continuation token is returned if there are more results than listed
+      here. You can optionally provide the token in the request payload to
+      load the additional results.
+additionalProperties: false
+required:
+  - roleIds
--- a/schemas/v1/list-role-ids-response.yml
+++ b/schemas/v1/list-role-ids-response.yml
@@ -1,14 +1,14 @@
 $schema:  http://json-schema.org/draft-06/schema#
 title:                      "Get Role Ids Response"
 description: |
-  Get a list of role IDs
+  If no limit is given, the roleIds of all roles are returned. Since this
+  list may become long, callers can use the `limit` and `continuationToken`
+  query arguments to page through the responses.
 type: object
 properties:
   roleIds:
     description: |
-      If no limit is given, the roleIds of all roles are returned. Since this
-      list may become long, callers can use the `limit` and `continuationToken`
-      query arguments to page through the responses.
+      A list of requested roleIds
     type:                 array
     items:              {type: string}
   continuationToken:
--- a/schemas/v1/list-role-ids-response.yml
+++ b/schemas/v1/list-role-ids-response.yml
@@ -1,7 +1,7 @@
 $schema:  http://json-schema.org/draft-06/schema#
-title:                      "Get Role Response"
+title:                      "Get Role Ids Response"
 description: |
-  Get all details about a role
+  Get roleIds of roles based on continuationToken and limit
 type: object
 properties:
   roleIds:
--- a/schemas/v1/list-role-ids-response.yml
+++ b/schemas/v1/list-role-ids-response.yml
@@ -6,12 +6,14 @@ type: object
 properties:
   roleIds:
     description: |
-      roleIds correspnding to the list of roles requested
+      If no limit is given, the roleIds of all roles are returned. Since this
+      list may become long, callers can use the `limit` and `continuationToken`
+      query arguments to page through the responses.
     type:                 array
     items:              {type: string}
   continuationToken:
     title:                "Continuation Token"
-    type:                 number
+    type:                 string
     description: |
       A continuation token is returned if there are more results than listed
       here. You can optionally provide the token in the request payload to
--- a/schemas/v1/list-role-ids-response.yml
+++ b/schemas/v1/list-role-ids-response.yml
@@ -1,7 +1,7 @@
 $schema:  http://json-schema.org/draft-06/schema#
 title:                      "Get Role Ids Response"
 description: |
-  Get roleIds of roles based on continuationToken and limit
+  Get a list of role IDs
 type: object
 properties:
   roleIds:
@@ -11,7 +11,7 @@ properties:
     items:              {type: string}
   continuationToken:
     title:                "Continuation Token"
-    type:                 string
+    type:                 number
     description: |
       A continuation token is returned if there are more results than listed
       here. You can optionally provide the token in the request payload to
--- a/src/v1.js
+++ b/src/v1.js
@@ -586,6 +586,48 @@ builder.declare({
   res.reply(roles.map(r => roleToJson(r, this)));
 });
 
+/** List role Ids **/
+builder.declare({
+  method:     'get',
+  route:      '/roleids/',
+  query: {
+    continuationToken: /./,
+    limit: /^[0-9]+$/,
+  },
+  name:       'listRoleIds  ',
+  input:      undefined,
+  output:     'list-role-ids-response.yml',
+  stability:  'stable',
+  title:      'List Role IDs',
+  description: [
+    'Get a list role IDs based on the limit and continuationToken',
+    'provided. If no limit is provided then all role IDs are returned.',
+  ].join('\n'),
+}, async function(req, res) {
+  let continuationToken = req.query.continuationToken || undefined;
+  let limit = parseInt(req.query.limit, 10);
+  let response = {};
+
+  // Load all roles
+  let roles = await this.Roles.get();
+
+  //slice the list of roles based on continuationToken and limit
+  if (continuationToken && limit) {
+    roles = roles.slice(continuationToken, limit+continuationToken);
+    response.continuationToken = limit+continuationToken;
+  } else if (limit) {
+    roles = roles.slice(0, limit);   //if no continuationToken is provided
+  }
+
+  //generate a list of roleIds corresponding to the selected roles
+  let roleIds = roles.map(r => r.roleId);
+
+  response.roleIds = roleIds;
+
+  res.reply(response);
+
+});
+
 /** Get role */
 builder.declare({
   method:     'get',
--- a/src/v1.js
+++ b/src/v1.js
@@ -607,9 +607,22 @@ builder.declare({
   ].join('\n'),
 }, async function(req, res) {
   let hashids = new Hashids();
-  let continuationToken = parseInt(req.query.continuationToken, 10) || undefined;
+  let continuationToken;
   let limit = parseInt(req.query.limit, 10) || undefined;
   let response = {};
+
+  // Assign the continuationToken
+  if (req.query.continuationToken) {
+    continuationToken = hashids.decode(req.query.continuationToken);
+    // If continuationToken is invalid
+    if(continuationToken.length === 0){
+      return res.reportError('InputError', 'Invalid continuationToken', {});
+    }
+    // Assign the decoded token value
+    continuationToken = continuationToken[0];
+  } else {
+    continuationToken = undefined;
+  }
 
   // Load all roles
   let roles = await this.Roles.get();
--- a/src/v1.js
+++ b/src/v1.js
@@ -6,6 +6,7 @@ const slugid = require('slugid');
 const _ = require('lodash');
 const signaturevalidator = require('./signaturevalidator');
 const ScopeResolver = require('./scoperesolver');
+const Hashids = require('hashids');
 
 /**
  * Helper to return a role as defined in the blob to one suitable for return.
@@ -600,10 +601,12 @@ builder.declare({
   stability:  'stable',
   title:      'List Role IDs',
   description: [
-    'Get a list of role IDs based on the limit and continuationToken',
-    'provided. If no limit is provided then all role IDs are returned.',
+    'If no limit is given, the roleIds of all roles are returned. Since this',
+    'list may become long, callers can use the `limit` and `continuationToken`',
+    'query arguments to page through the responses.',
   ].join('\n'),
 }, async function(req, res) {
+  let hashids = new Hashids();
   let continuationToken = parseInt(req.query.continuationToken, 10);
   let limit = parseInt(req.query.limit, 10);
   let response = {};
@@ -618,14 +621,14 @@ builder.declare({
     continuationToken = limit+continuationToken;
 
     if (continuationToken < length) {
-      response.continuationToken = continuationToken;
+      response.continuationToken = hashids.encode(continuationToken);
     }
   } else if (limit) {
     roles = roles.slice(0, limit);   // If no continuationToken is provided
     continuationToken = limit;
 
     if (continuationToken < length) {
-      response.continuationToken = continuationToken;
+      response.continuationToken = hashids.encode(continuationToken);
     }
   }
 
--- a/src/v1.js
+++ b/src/v1.js
@@ -607,8 +607,8 @@ builder.declare({
   ].join('\n'),
 }, async function(req, res) {
   let hashids = new Hashids();
-  let continuationToken = parseInt(req.query.continuationToken, 10);
-  let limit = parseInt(req.query.limit, 10);
+  let continuationToken = parseInt(req.query.continuationToken, 10) || undefined;
+  let limit = parseInt(req.query.limit, 10) || undefined;
   let response = {};
 
   // Load all roles
@@ -617,18 +617,18 @@ builder.declare({
 
   // Slice the list of roles based on continuationToken and limit
   if (continuationToken && limit) {
-    roles = roles.slice(continuationToken, limit+continuationToken);
-    continuationToken = limit+continuationToken;
+    roles = roles.slice(continuationToken, limit + continuationToken);
+    continuationToken = limit + continuationToken;
 
     if (continuationToken < length) {
-      response.continuationToken = hashids.encode(continuationToken);
+      response.continuationToken = hashids.encode(continuationToken, 10);
     }
   } else if (limit) {
     roles = roles.slice(0, limit);   // If no continuationToken is provided
     continuationToken = limit;
 
     if (continuationToken < length) {
-      response.continuationToken = hashids.encode(continuationToken);
+      response.continuationToken = hashids.encode(continuationToken, 10);
     }
   }
 
--- a/src/v1.js
+++ b/src/v1.js
@@ -594,39 +594,47 @@ builder.declare({
     continuationToken: /./,
     limit: /^[0-9]+$/,
   },
-  name:       'listRoleIds  ',
+  name:       'listRoleIds',
   input:      undefined,
   output:     'list-role-ids-response.yml',
   stability:  'stable',
   title:      'List Role IDs',
   description: [
-    'Get a list role IDs based on the limit and continuationToken',
+    'Get a list of role IDs based on the limit and continuationToken',
     'provided. If no limit is provided then all role IDs are returned.',
   ].join('\n'),
 }, async function(req, res) {
-  let continuationToken = req.query.continuationToken || undefined;
+  let continuationToken = parseInt(req.query.continuationToken, 10);
   let limit = parseInt(req.query.limit, 10);
   let response = {};
 
   // Load all roles
   let roles = await this.Roles.get();
+  let length = roles.length;
 
-  //slice the list of roles based on continuationToken and limit
+  // Slice the list of roles based on continuationToken and limit
   if (continuationToken && limit) {
     roles = roles.slice(continuationToken, limit+continuationToken);
-    response.continuationToken = limit+continuationToken;
+    continuationToken = limit+continuationToken;
+
+    if (continuationToken < length) {
+      response.continuationToken = continuationToken;
+    }
   } else if (limit) {
-    roles = roles.slice(0, limit);   //if no continuationToken is provided
-    response.continuationToken = limit;
+    roles = roles.slice(0, limit);   // If no continuationToken is provided
+    continuationToken = limit;
+
+    if (continuationToken < length) {
+      response.continuationToken = continuationToken;
+    }
   }
 
-  //generate a list of roleIds corresponding to the selected roles
+  // Generate a list of roleIds corresponding to the selected roles
   let roleIds = roles.map(r => r.roleId);
 
   response.roleIds = roleIds;
 
   res.reply(response);
-
 });
 
 /** Get role */
--- a/src/v1.js
+++ b/src/v1.js
@@ -617,6 +617,7 @@ builder.declare({
     response.continuationToken = limit+continuationToken;
   } else if (limit) {
     roles = roles.slice(0, limit);   //if no continuationToken is provided
+    response.continuationToken = limit;
   }
 
   //generate a list of roleIds corresponding to the selected roles
--- a/src/v1.js
+++ b/src/v1.js
@@ -615,7 +615,7 @@ builder.declare({
   if (req.query.continuationToken) {
     continuationToken = hashids.decode(req.query.continuationToken);
     // If continuationToken is invalid
-    if(continuationToken.length === 0){
+    if (continuationToken.length === 0) {
       return res.reportError('InputError', 'Invalid continuationToken', {});
     }
     // Assign the decoded token value
--- a/test/role_test.js
+++ b/test/role_test.js
@@ -136,20 +136,21 @@ helper.secrets.mockSuite(helper.suiteName(__filename), ['app', 'azure'], functio
 
   test('listRoles', async () => {
     let roles = await helper.apiClient.listRoles();
-    assert(roles.some(role => role.roleId === 'thing-id:' + clientId));
+    assert(roles.some(role => role.roleId === `thing-id:${clientId}`));
   });
 
   test('listRoleIds', async () => {
+    // Clear existing roles
     await helper.Roles.modify((roles) => roles.splice(0));
 
     // Create 4 dummy roles
-    await helper.apiClient.createRole('thing-id:' + clientId, {
+    await helper.apiClient.createRole(`thing-id:${clientId}`, {
       description: 'test role',
       scopes: ['dummy-scope-1', 'auth:create-role:*', 'dummy-scope-2'],
     });
     for (let i=0;i<3;i++) {
-      let tempClientId = clientId + i;
-      await helper.apiClient.createRole('thing-id:' + tempClientId, {
+      let tempRoleId = `${clientId}${i}`;
+      await helper.apiClient.createRole(`thing-id:${clientId}`, {
         description: 'test role',
         scopes: ['dummy-scope-1', 'auth:create-role:*', 'dummy-scope-2'],
       });
@@ -157,7 +158,7 @@ helper.secrets.mockSuite(helper.suiteName(__filename), ['app', 'azure'], functio
 
     let result = await helper.apiClient.listRoleIds();
 
-    assert(result.roleIds.some(roleId => roleId === 'thing-id:' + clientId));
+    assert(result.roleIds.some(roleId => roleId === `thing-id:${clientId}`));
     assert(result.roleIds.length === 4);
   });
 
@@ -182,7 +183,7 @@ helper.secrets.mockSuite(helper.suiteName(__filename), ['app', 'azure'], functio
     }
 
     assume(roleIds.sort()).to.deeply.equal(allRoleIds.roleIds.sort());
-    assume(count).to.be.greaterThan(0);
+    assume(count).to.be.greaterThan(1);
   });
 
   test('updateRole with a **-scope', async () => {
@@ -194,23 +195,24 @@ helper.secrets.mockSuite(helper.suiteName(__filename), ['app', 'azure'], functio
   });
 
   test('updateRole (add scope)', async () => {
+    // Clearing existing roles
     await helper.Roles.modify((roles) => roles.splice(0));
 
-    // Generating dummy roles
-    await helper.apiClient.createRole('thing-id:' + clientId, {
+    // Generating dummy roles for the test
+    await helper.apiClient.createRole(`thing-id:${clientId}`, {
       description: 'test role',
       scopes: ['dummy-scope-1', 'auth:create-role:*', 'dummy-scope-2'],
     });
-    await helper.apiClient.createRole('thing-id:' + clientId.slice(0, 11) + '*', {
+    await helper.apiClient.createRole(`thing-id:${clientId.slice(0, 11)}*`, {
       description: 'test role',
       scopes: ['dummy-scope-2'],
     });
 
-    let r1 = await helper.apiClient.role('thing-id:' + clientId);
+    let r1 = await helper.apiClient.role(`thing-id:${clientId}`);
 
     await testing.sleep(100);
 
-    let r2 = await helper.apiClient.updateRole('thing-id:' + clientId, {
+    let r2 = await helper.apiClient.updateRole(`thing-id:${clientId}`, {
       description: 'test role',
       scopes: ['dummy-scope-1', 'auth:create-role:*', 'dummy-scope-3'],
     });
@@ -220,9 +222,9 @@ helper.secrets.mockSuite(helper.suiteName(__filename), ['app', 'azure'], functio
     );
     helper.checkNextMessage('role-updated', m => assert.equal(m.payload.roleId, `thing-id:${clientId}`));
 
-    let role = await helper.apiClient.role('thing-id:' + clientId);
+    let role = await helper.apiClient.role(`thing-id:${clientId}`);
     assume(role.expandedScopes.sort()).deep.equals([
-      'assume:thing-id:' + clientId,
+      `assume:thing-id:${clientId}`,
       'dummy-scope-1',
       'auth:create-role:*',
       'dummy-scope-2', // from role thing-id:<clientId[:11]>*
--- a/test/role_test.js
+++ b/test/role_test.js
@@ -138,6 +138,29 @@ helper.secrets.mockSuite(helper.suiteName(__filename), ['app', 'azure'], functio
     assert(roles.some(role => role.roleId === 'thing-id:' + clientId));
   });
 
+  test('listRoleIds', async () => {
+    let result = await helper.apiClient.listRoleIds();
+    assert(result.roleIds.some(roleId => roleId === 'thing-id:' + clientId));
+  });
+
+  test('listRoleIds (limit, [continuationToken])', async () => {
+    let roleIds = [];
+    let query = {limit:1};
+    let allRoleIds = await helper.apiClient.listRoleIds();
+    
+    while (true) {
+      let result = await helper.apiClient.listRoleIds(query);
+      assume(result.roleIds.length).to.be.lessThan(2);
+      query.continuationToken = result.continuationToken;
+      roleIds = roleIds.concat(result.roleIds);
+      if (!query.continuationToken) {
+        break;
+      }
+    }
+
+    assume(roleIds).to.deeply.equal(allRoleIds.roleIds);
+  });
+
   test('updateRole with a **-scope', async () => {
     await helper.apiClient.updateRole('thing-id:' + clientId, {
       description: 'other',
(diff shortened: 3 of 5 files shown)


schemas/v1/list-role-ids-response.yml
```diff
@@ -0,0 +1,21 @@
+$schema:  http://json-schema.org/draft-06/schema#
+title:                      "Get Role Ids Response"
+description: |
+  Get roleIds of roles based on continuationToken and limit
```
helfi92: "continuationToken and limit" may not make sense to people not familiar with the internals of Taskcluster. What do you think about "List of role IDs"?

djmitche:
Regarding adding tests, have a look at `test/role_test.js`, which tests the existing role-related API methods.  You should be able to copy-and-modify the tests in that file to suit your new method.  There are some tricks to testing continuation/limit in the tests for listClients, which are in `client_test.js` (I think - a quick `git grep` will find them).

You can run just the one test file quickly with `mocha test/role_test.js` if that helps you go faster :)

src/v1.js
```diff
@@ -586,6 +586,49 @@ builder.declare({
...
+    continuationToken: /./,
+    limit: /^[0-9]+$/,
+  },
+  name:       'listRoleIds  ',
+  input:      undefined,
+  output:     'list-role-ids-response.yml',
+  stability:  'stable',
+  title:      'List Role IDs',
+  description: [
+    'Get a list role IDs based on the limit and continuationToken',
```
helfi92: `s/list role IDs/list of role IDs/`.

src/v1.js
```diff
@@ -586,6 +586,49 @@ builder.declare({
...
+    response.continuationToken = limit;
+  }
+
+  //generate a list of roleIds corresponding to the selected roles
+  let roleIds = roles.map(r => r.roleId);
+
+  response.roleIds = roleIds;
+
+  res.reply(response);
+
```
helfi92: Remove extra line.

src/v1.js
```diff
@@ -586,6 +586,49 @@ builder.declare({
...
+  ].join('\n'),
+}, async function(req, res) {
+  let continuationToken = req.query.continuationToken || undefined;
+  let limit = parseInt(req.query.limit, 10);
+  let response = {};
+
+  // Load all roles
+  let roles = await this.Roles.get();
+
+  //slice the list of roles based on continuationToken and limit
```
helfi92: Add a space after `//`.

src/v1.js
```diff
@@ -586,6 +586,49 @@ builder.declare({
...
+  if (continuationToken && limit) {
+    roles = roles.slice(continuationToken, limit+continuationToken);
+    response.continuationToken = limit+continuationToken;
+  } else if (limit) {
+    roles = roles.slice(0, limit);   //if no continuationToken is provided
+    response.continuationToken = limit;
+  }
+
+  //generate a list of roleIds corresponding to the selected roles
```
helfi92: Add a space after `//`.

src/v1.js
OjaswinM: My bad, haha. Also, should I remove limit and continuationToken from here as well?

schemas/v1/list-role-ids-response.yml
OjaswinM: Right, that makes more sense!

src/v1.js
```diff
@@ -586,6 +586,49 @@ builder.declare({
...
+  output:     'list-role-ids-response.yml',
+  stability:  'stable',
+  title:      'List Role IDs',
+  description: [
+    'Get a list role IDs based on the limit and continuationToken',
+    'provided. If no limit is provided then all role IDs are returned.',
+  ].join('\n'),
+}, async function(req, res) {
+  let continuationToken = req.query.continuationToken || undefined;
+  let limit = parseInt(req.query.limit, 10);
```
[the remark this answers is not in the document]
OjaswinM: As it is, if limit is not passed then a list of roleIDs of all the roles are returned ([comment 5](https://bugzilla.mozilla.org/show_bug.cgi?id=1455130#c5)). Should I keep this behavior or use a default?

src/v1.js
```diff
@@ -586,6 +586,49 @@ builder.declare({
...
+  input:      undefined,
+  output:     'list-role-ids-response.yml',
+  stability:  'stable',
+  title:      'List Role IDs',
+  description: [
+    'Get a list role IDs based on the limit and continuationToken',
+    'provided. If no limit is provided then all role IDs are returned.',
+  ].join('\n'),
+}, async function(req, res) {
+  let continuationToken = req.query.continuationToken || undefined;
```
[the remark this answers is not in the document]
OjaswinM: Oh right, I'll add that! As for hashids, yes I plan on doing that in future commits :)

schemas/v1/list-role-ids-response.yml
```diff
@@ -0,0 +1,21 @@
+$schema:  http://json-schema.org/draft-06/schema#
+title:                      "Get Role Ids Response"
+description: |
+  Get a list of role IDs
```
djmitche: Let's add some text here:

> If no limit is given, the roleIds of all roles are returned.  Since this list may become long, callers can use the `limit` and `continuationToken` query arguments to page through the responses.

test/role_test.js
```diff
@@ -138,6 +138,29 @@ helper.secrets.mockSuite(helper.suiteName(__filename), ['app', 'azure'], functio
     assert(roles.some(role => role.roleId === 'thing-id:' + clientId));
   });
 
+  test('listRoleIds', async () => {
+    let result = await helper.apiClient.listRoleIds();
+    assert(result.roleIds.some(roleId => roleId === 'thing-id:' + clientId));
```
djmitche: It'd be good to assert that you get the whole list here.

test/role_test.js
```diff
@@ -138,6 +138,29 @@ helper.secrets.mockSuite(helper.suiteName(__filename), ['app', 'azure'], functio
...
+      let result = await helper.apiClient.listRoleIds(query);
+      assume(result.roleIds.length).to.be.lessThan(2);
+      query.continuationToken = result.continuationToken;
+      roleIds = roleIds.concat(result.roleIds);
+      if (!query.continuationToken) {
+        break;
+      }
+    }
+
+    assume(roleIds).to.deeply.equal(allRoleIds.roleIds);
```
djmitche: Can you add a check here that the loop iterated more than once?

test/role_test.js
OjaswinM: How should I do that? I'll need the actual length of the list right?

djmitche: Hm, good question.  How many roles are there at this point?  It looks like this test suite does not clear the list of roles between test executions, so what's left at this point is whatever the previous tests have created.

You could reset the roles to an empty list with
```js
await this.Roles.modify(roles => ([]));
```
and then fill it with a set of roles by calling createRole a few times in a loop.  That would, at least, give you a stable set of roles to test against.

test/role_test.js
djmitche: One more thing: we do not guarantee anything about the order of roles, so it might be good to add `.sort()` to both `roleIds` and `allRoleIds.roleIds` here.

test/role_test.js
OjaswinM: this.Roles seems to be undefined. I'm getting this

`TypeError: Cannot read property 'modify' of undefined`

djmitche: Oops, sorry, `helper.Roles`

djmitche: Ah, good fix.

OjaswinM:
All the changes are made including hashids. Deleting and adding dummy roles in listRoleIds()'s test broke the updateRole(add scope) test as it made some assumptions about the existing roles so I generated some dummy roles for that test as well. Kindly review the code :)

src/v1.js
```diff
@@ -586,6 +587,59 @@ builder.declare({
...
+  stability:  'stable',
+  title:      'List Role IDs',
+  description: [
+    'If no limit is given, the roleIds of all roles are returned. Since this',
+    'list may become long, callers can use the `limit` and `continuationToken`',
+    'query arguments to page through the responses.',
+  ].join('\n'),
+}, async function(req, res) {
+  let hashids = new Hashids();
+  let continuationToken = parseInt(req.query.continuationToken, 10);
```
helfi92: If the continuation token is not defined, then this will return `NaN`. Let's do:

```js
let continuationToken = parseInt(req.query.continuationToken, 10) || undefined;
```

src/v1.js
```diff
@@ -586,6 +587,59 @@ builder.declare({
...
+  let limit = parseInt(req.query.limit, 10);
+  let response = {};
+
+  // Load all roles
+  let roles = await this.Roles.get();
```
helfi92: Silly question but why are not doing `Roles.scan({}, opts);`?

src/v1.js
```diff
@@ -586,6 +587,59 @@ builder.declare({
...
+  let length = roles.length;
+
+  // Slice the list of roles based on continuationToken and limit
+  if (continuationToken && limit) {
+    roles = roles.slice(continuationToken, limit+continuationToken);
```
helfi92: Add spaces between the `+` to make it more breathable :)

src/v1.js
```diff
@@ -586,6 +587,59 @@ builder.declare({
...
+    continuationToken = limit+continuationToken;
```
helfi92: Add spaces between `+`:

```js
continuationToken = limit + continuationToken;

```

src/v1.js
```diff
@@ -586,6 +587,59 @@ builder.declare({
...
+
+    if (continuationToken < length) {
+      response.continuationToken = hashids.encode(continuationToken);
```
helfi92: You can add more padding to the encoded value with the `encode` second parameter:

```js
response.continuationToken = hashids.encode(continuationToken, 10);
```

src/v1.js
```diff
@@ -586,6 +587,59 @@ builder.declare({
...
+    }
+  } else if (limit) {
+    roles = roles.slice(0, limit);   // If no continuationToken is provided
+    continuationToken = limit;
+
+    if (continuationToken < length) {
+      response.continuationToken = hashids.encode(continuationToken);
```
helfi92: Similar to the previous comment, you can add more padding:

```js
response.continuationToken = hashids.encode(continuationToken, 10);
```

test/role_test.js
```diff
@@ -138,6 +139,52 @@ helper.secrets.mockSuite(helper.suiteName(__filename), ['app', 'azure'], functio
     assert(roles.some(role => role.roleId === 'thing-id:' + clientId));
   });
 
+  test('listRoleIds', async () => {
+    await helper.Roles.modify((roles) => roles.splice(0));
```
helfi92: What are we trying to do here? `roles.splice(0)` destroys the old array and replaces it with a copy. Perhaps add a comment above to make it clearer :)

test/role_test.js
```diff
@@ -138,6 +139,52 @@ helper.secrets.mockSuite(helper.suiteName(__filename), ['app', 'azure'], functio
...
+
+    // Create 4 dummy roles
+    await helper.apiClient.createRole('thing-id:' + clientId, {
```
helfi92: Use template strings instead of `+` when you can:

``` `thing-id:${clientId}` ```

File: test/role_test.js [hunk 19]
```diff
@@ -138,6 +139,52 @@ helper.secrets.mockSuite(helper.suiteName(__filename), ['app', 'azure'], functio
...
+      description: 'test role',
+      scopes: ['dummy-scope-1', 'auth:create-role:*', 'dummy-scope-2'],
+    });
+    for (let i=0;i<3;i++) {
+      let tempClientId = clientId + i;
+      await helper.apiClient.createRole('thing-id:' + tempClientId, {
```
helfi92: Use template strings for the first argument.

test/role_test.js
```diff
@@ -138,6 +139,52 @@ helper.secrets.mockSuite(helper.suiteName(__filename), ['app', 'azure'], functio
...
+  test('listRoleIds', async () => {
+    await helper.Roles.modify((roles) => roles.splice(0));
+
+    // Create 4 dummy roles
+    await helper.apiClient.createRole('thing-id:' + clientId, {
+      description: 'test role',
+      scopes: ['dummy-scope-1', 'auth:create-role:*', 'dummy-scope-2'],
+    });
+    for (let i=0;i<3;i++) {
+      let tempClientId = clientId + i;
```
helfi92: Given that we are dealing with roles per say, should we rename this to `tempRoleId`?

test/role_test.js
```diff
@@ -138,6 +139,52 @@ helper.secrets.mockSuite(helper.suiteName(__filename), ['app', 'azure'], functio
...
+      await helper.apiClient.createRole('thing-id:' + tempClientId, {
+        description: 'test role',
+        scopes: ['dummy-scope-1', 'auth:create-role:*', 'dummy-scope-2'],
+      });
+    }
+
+    let result = await helper.apiClient.listRoleIds();
+
+    assert(result.roleIds.some(roleId => roleId === 'thing-id:' + clientId));
```
helfi92: Use template strings instead of `+` when you can.

test/role_test.js
```diff
@@ -138,6 +139,52 @@ helper.secrets.mockSuite(helper.suiteName(__filename), ['app', 'azure'], functio
...
+    let allRoleIds = {};
+    let count=0;
+
+    allRoleIds = await helper.apiClient.listRoleIds();
+
+    while (true) {
+      let result = await helper.apiClient.listRoleIds(query);
+      assume(result.roleIds.length).to.be.lessThan(2);
+      query.continuationToken = hashids.decode(result.continuationToken)[0];
+      roleIds = roleIds.concat(result.roleIds);
```
helfi92: Instead of reassigning `roleIds`, I think you can just do `roleIds.push(result.roleIds);`.

test/role_test.js
```diff
@@ -138,6 +139,52 @@ helper.secrets.mockSuite(helper.suiteName(__filename), ['app', 'azure'], functio
...
+      count++;
+      if (!query.continuationToken) {
+        break;
+      }
+    }
+
+    assume(roleIds.sort()).to.deeply.equal(allRoleIds.roleIds.sort());
+    assume(count).to.be.greaterThan(0);
```
helfi92: This will always be true since `count++` will always be executed at least once before reaching the `break` statement. Maybe we should do `greaterThan(1)`?

src/v1.js
djmitche: Roles is a single blob in Azure blob storage, rather than an Azure table with multiple rows.

With parameterized roles, there are combinations of roles that are invalid.  We need to prevent such combinations from being written to the role storage.  It's not possible to make atomic updates to a entire Azure table, so if we were using table storage, race conditions would permit writing invalid combinations to the table.  So, we switched to a single blob that we can read and write atomically.

test/role_test.js
djmitche: pushing an array would create a sub-array..

helfi92: You're right.

src/v1.js
OjaswinM: Will do! I thought padding was not required so didn't add it initially.

test/role_test.js
OjaswinM: This is to clear the roles created by the other tests. I'll add a comment explaining it :)

test/role_test.js
OjaswinM: Right, that makes more sense

OjaswinM:
I've made the requested changes, kindly review :)

src/v1.js
```diff
@@ -586,6 +587,59 @@ builder.declare({
...
+  stability:  'stable',
+  title:      'List Role IDs',
+  description: [
+    'If no limit is given, the roleIds of all roles are returned. Since this',
+    'list may become long, callers can use the `limit` and `continuationToken`',
+    'query arguments to page through the responses.',
+  ].join('\n'),
+}, async function(req, res) {
+  let hashids = new Hashids();
+  let continuationToken = parseInt(req.query.continuationToken, 10) || undefined;
```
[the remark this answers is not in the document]
OjaswinM: Hey, so firstly, what will the corect way of returning the error be? Secondly, under what case should we consider the continuationToken invalid? I can only think of two possibilities

1. It's value is greater than the total length of the list of roles.

2. It's passed without a limit. I'm not sure if this should be considered invalid as such but our code doesn't account for this case currently.

Thoughts?

test/role_test.js
```diff
@@ -135,7 +136,54 @@ helper.secrets.mockSuite(helper.suiteName(__filename), ['app', 'azure'], functio
...
+    let query = {limit:1};
+    let allRoleIds = {};
+    let count=0;
+
+    allRoleIds = await helper.apiClient.listRoleIds();
+
+    while (true) {
+      let result = await helper.apiClient.listRoleIds(query);
+      assume(result.roleIds.length).to.be.lessThan(2);
+      query.continuationToken = hashids.decode(result.continuationToken)[0];
```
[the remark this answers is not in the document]
OjaswinM: Oh right, this is a silly mistake on my part. I'll correct it asap :)

src/v1.js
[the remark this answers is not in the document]
djmitche: The kind of "invalid" I was thinking of was when `hashids.decode()` fails to decode it -- for example `?continuationToken=FOOBAR&limit=10`.

I think passing continuationToken without a limit is OK -- it just means requesting the default limit.

OjaswinM:
I've made the requested changes, kindly check :)

djmitche:
@helfi92 maybe a good next-step would be adding support for this method to tc-web-server?

OjaswinM:
Oh, my bad. I will keep that in mind next time :)

OjaswinM:
Sure, I would love to work on it! Many thanks to you and Dustin for guiding me through this :)

helfi92:
@djmitche I am trying to update `taskcluster-client` via `bin/update-apis.js update` however it's generating a weird diff (strings becoming `undefined` and urls changing). Below are some of the interesting parts of the diff:

```diff
diff --git a/README.md b/README.md
index 4b67094..1b83aae 100644
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ test('test the thing', async function() {
 ### Methods in `taskcluster.Auth`
 ```js
 // Create Auth client instance:
-//  - https://auth.taskcluster.net/v1/
+//  - undefined
 var auth = new taskcluster.Auth(options);
 // Create Github client instance:
-//  - https://github.taskcluster.net/v1/
+//  - undefined
 var github = new taskcluster.Github(options);
 // Create Hooks client instance:
-//  - https://hooks.taskcluster.net/v1/
+//  - undefined
 // Create Index client instance:
-//  - https://index.taskcluster.net/v1/
+//  - undefined
 // Create Login client instance:
-//  - https://login.taskcluster.net/v1
+//  - undefined
 var login = new taskcluster.Login(options);
 // Create Queue client instance:
-//  - https://queue.taskcluster.net/v1/
+//  - undefined
 var queue = new taskcluster.Queue(options);
 diff --git a/src/apis.js b/src/apis.js
index 1efa755..83cd46e 100644
--- a/src/apis.js
+++ b/src/apis.js
@@ -2,8 +2,8 @@
 module.exports = {
   "Auth": {
     "reference": {
-      "$schema": "http://schemas.taskcluster.net/base/v1/api-reference.json#",
-      "baseUrl": "https://auth.taskcluster.net/v1/",
+      "$schema": "/schemas/common/api-reference-v0.json#",
+      "apiVersion": "v1",
@@ -530,15 +530,15 @@ module.exports = {
         }
       ],
       "serviceName": "auth",
-      "title": "Authentication API",
-      "version": 0
+      "title": "Authentication API"
     },
     "referenceUrl": "https://references.taskcluster.net/auth/v1/api.json"
   },
   "AuthEvents": {
     "reference": {
-      "$schema": "http://schemas.taskcluster.net/base/v1/exchanges-reference.json#",
-      "description": "The auth service, typically available at `auth.taskcluster.net`\nis responsible for storing credentials, managing assignment of scopes,\nand validation of request signatures from other services.\n\nThese exchanges provides notifications when credentials or roles are\nupdated. This is mostly so that multiple instances of the auth service\ncan purge their caches and synchronize state. But you are of course\nwelcome to use these for other purposes, monitoring changes for example.",
+      "$schema": "/schemas/common/exchanges-reference-v0.json#",
+      "apiVersion": "v1",
+      "description": "The auth service is responsible for storing credentials, managing\nassignment of scopes, and validation of request signatures from other\nservices.\n\nThese exchanges provides notifications when credentials or roles are\nupdated. This is mostly so that multiple instances of the auth service\ncan purge their caches and synchronize state. But you are of course\nwelcome to use these for other purposes, monitoring changes for example.",
       "entries": [
         {
           "description": "Message that a new client has been created.",
@@ -639,14 +639,14 @@ module.exports = {
       ],
       "exchangePrefix": "exchange/taskcluster-auth/v1/",
       "serviceName": "auth",
-      "title": "Auth Pulse Exchanges",
-      "version": 0
+      "title": "Auth Pulse Exchanges"
     },
     "referenceUrl": "https://references.taskcluster.net/auth/v1/exchanges.json"
   },
   "AwsProvisioner": {
     "reference": {
       "$schema": "http://schemas.taskcluster.net/base/v1/api-reference.json#",
+      "apiVersion": "v1",
       "baseUrl": "https://aws-provisioner.taskcluster.net/v1",
       "description": "The AWS Provisioner is responsible for provisioning instances on EC2 for use in\nTaskcluster.  The provisioner maintains a set of worker configurations which\ncan be managed with an API that is typically available at\naws-provisioner.taskcluster.net/v1.  This API can also perform basic instance\nmanagement tasks in addition to maintaining the internal state of worker type\nconfiguration information.\n\nThe Provisioner runs at a configurable interval.  Each iteration of the\nprovisioner fetches a current copy the state that the AWS EC2 api reports.  In\neach iteration, we ask the Queue how many tasks are pending for that worker\ntype.  Based on the number of tasks pending and the scaling ratio, we may\nsubmit requests for new instances.  We use pricing information, capacity and\nutility factor information to decide which instance type in which region would\nbe the optimal configuration.\n\nEach EC2 instance type will declare a capacity and utility factor.  Capacity is\nthe number of tasks that a given machine is capable of running concurrently.\nUtility factor is a relative measure of performance between two instance types.\nWe multiply the utility factor by the spot price to compare instance types and\nregions when making the bidding choices.\n\nWhen a new EC2 instance is instantiated, its user data contains a token in\n`securityToken` that can be used with the `getSecret` method to retrieve\nthe worker's credentials and any needed passwords or other restricted\ninformation.  The worker is responsible for deleting the secret after\nretrieving it, to prevent dissemination of the secret to other proceses\nwhich can read the instance user data.\n",
       "entries": [
@@ -886,7 +886,7 @@ module.exports = {
           "type": "function"
         }
       ],
-      "name": "aws-provisioner",
+      "serviceName": "aws-provisioner",
       "title": "AWS Provisioner API Documentation",
       "version": 0
     },
@@ -1178,7 +1176,7 @@ module.exports = {
           "description": "Return a list of possible prices for EC2",
           "method": "get",
           "name": "getPrices",
-          "output": "http://schemas.taskcluster.net/ec2-manager/v1/prices.json#",
+          "output": "v1/prices.json#",
           "query": [
           ],
           "route": "/prices",
@@ -1190,10 +1188,10 @@ module.exports = {
           "args": [
           ],
           "description": "Return a list of possible prices for EC2",
-          "input": "http://schemas.taskcluster.net/ec2-manager/v1/prices-request.json#",
+          "input": "v1/prices-request.json#",
           "method": "post",
           "name": "getSpecificPrices",
-          "output": "http://schemas.taskcluster.net/ec2-manager/v1/prices.json#",
+          "output": "v1/prices.json#",
           "query": [
           ],
           "route": "/prices",
@@ -1207,7 +1205,7 @@ module.exports = {
           "description": "Give some basic stats on the health of our EC2 account",
           "method": "get",
           "name": "getHealth",
-          "output": "http://schemas.taskcluster.net/ec2-manager/v1/health.json#",
+          "output": "v1/health.json#",
           "query": [
           ],
           "route": "/health",
@@ -1221,7 +1219,7 @@ module.exports = {
           "description": "Return a list of recent errors encountered",
           "method": "get",
           "name": "getRecentErrors",
-          "output": "http://schemas.taskcluster.net/ec2-manager/v1/errors.json#",
+          "output": "v1/errors.json#",
           "query": [
           ],
@@ -1389,9 +1334,9 @@ module.exports = {
   },
   "Github": {
     "reference": {
-      "$schema": "http://schemas.taskcluster.net/base/v1/api-reference.json#",
-      "baseUrl": "https://github.taskcluster.net/v1/",
-      "description": "The github service, typically available at\n`github.taskcluster.net`, is responsible for publishing pulse\nmessages in response to GitHub events.\n\nThis document describes the API end-point for consuming GitHub\nweb hooks, as well as some useful consumer APIs.\n\nWhen Github forbids an action, this service returns an HTTP 403\nwith code ForbiddenByGithub.",
+      "$schema": "/schemas/common/api-reference-v0.json#",
+      "apiVersion": "v1",
+      "description": "The github service is responsible for creating tasks in reposnse\nto GitHub events, and posting results to the GitHub UI.\n\nThis document describes the API end-point for consuming GitHub\nweb hooks, as well as some useful consumer APIs.\n\nWhen Github forbids an action, this service returns an HTTP 403\nwith code ForbiddenByGithub.",
       "entries": [
         {
           "args": [
@@ -1524,14 +1469,14 @@ module.exports = {
         }
       ],
       "serviceName": "github",
-      "title": "Taskcluster GitHub API Documentation",
-      "version": 0
+      "title": "Taskcluster GitHub API Documentation"
     },
     "referenceUrl": "https://references.taskcluster.net/github/v1/api.json"
   },
   "GithubEvents": {
     "reference": {
-      "$schema": "http://schemas.taskcluster.net/base/v1/exchanges-reference.json#",
+      "$schema": "/schemas/common/exchanges-reference-v0.json#",
+      "apiVersion": "v1",
       "description": "The github service publishes a pulse\nmessage for supported github events, translating Github webhook\nevents into pulse messages.\n\nThis document describes the exchange offered by the taskcluster\ngithub service",
       "entries": [
         {
@@ -1960,15 +1932,14 @@ module.exports = {
         }
       ],
       "serviceName": "index",
-      "title": "Task Index API Documentation",
-      "version": 0
+      "title": "Task Index API Documentation"
     },
     "referenceUrl": "https://references.taskcluster.net/index/v1/api.json"
   },
   "Login": {
     "reference": {
-      "$schema": "http://schemas.taskcluster.net/base/v1/api-reference.json#",
-      "baseUrl": "https://login.taskcluster.net/v1",
+      "$schema": "/schemas/common/api-reference-v0.json#",
+      "apiVersion": "v1",
       "description": "The Login service serves as the interface between external authentication\nsystems and Taskcluster credentials.",
       "entries": [
         {
@@ -2001,15 +1972,14 @@ module.exports = {
         }
       ],
       "serviceName": "login",
-      "title": "Login API",
-      "version": 0
+      "title": "Login API"
     },
     "referenceUrl": "https://references.taskcluster.net/login/v1/api.json"
   },
   "Notify": {
     "reference": {
-      "$schema": "http://schemas.taskcluster.net/base/v1/api-reference.json#",
-      "baseUrl": "https://notify.taskcluster.net/v1/",
+      "$schema": "/schemas/common/api-reference-v0.json#",
+      "apiVersion": "v1",
       "description": "The notification service, typically available at `notify.taskcluster.net`\nlistens for tasks with associated notifications and handles requests to\nsend emails and post pulse messages.",
       "entries": [
         {
@@ -2076,8 +2046,7 @@ module.exports = {
         }
       ],
       "serviceName": "notify",
-      "title": "Notification Service",
-      "version": 0
+      "title": "Notification Service"
     },
     "referenceUrl": "https://references.taskcluster.net/notify/v1/api.json"
   },
@@ -2161,9 +2130,9 @@ module.exports = {
   },
   "PurgeCache": {
     "reference": {
-      "$schema": "http://schemas.taskcluster.net/base/v1/api-reference.json#",
-      "baseUrl": "https://purge-cache.taskcluster.net/v1/",
-      "description": "The purge-cache service, typically available at\n`purge-cache.taskcluster.net`, is responsible for publishing a pulse\nmessage for workers, so they can purge cache upon request.\n\nThis document describes the API end-point for publishing the pulse\nmessage. This is mainly intended to be used by tools.",
+      "$schema": "/schemas/common/api-reference-v0.json#",
+      "apiVersion": "v1",
+      "description": "The purge-cache service is responsible for publishing a pulse\nmessage for workers, so they can purge cache upon request.\n\nThis document describes the API end-point for publishing the pulse\nmessage. This is mainly intended to be used by tools.",
       "entries": [
         {
           "args": [
@@ -2230,8 +2199,7 @@ module.exports = {
         }
       ],
       "serviceName": "purge-cache",
-      "title": "Purge Cache API",
-      "version": 0
+      "title": "Purge Cache API"
     },
     "referenceUrl": "https://references.taskcluster.net/purge-cache/v1/api.json"
   },
@@ -2279,9 +2247,9 @@ module.exports = {
   },
   "Queue": {
     "reference": {
-      "$schema": "http://schemas.taskcluster.net/base/v1/api-reference.json#",
-      "baseUrl": "https://queue.taskcluster.net/v1/",
-      "description": "The queue, typically available at `queue.taskcluster.net`, is responsible\nfor accepting tasks and track their state as they are executed by\nworkers. In order ensure they are eventually resolved.\n\nThis document describes the API end-points offered by the queue. These \nend-points targets the following audience:\n * Schedulers, who create tasks to be executed,\n * Workers, who execute tasks, and\n * Tools, that wants to inspect the state of a task.",
+      "$schema": "/schemas/common/api-reference-v0.json#",
+      "apiVersion": "v1",
+      "description": "The queue service is responsible for accepting tasks and track their state\nas they are executed by workers. In order ensure they are eventually\nresolved.\n\nThis document describes the API end-points offered by the queue. These \nend-points targets the following audience:\n * Schedulers, who create tasks to be executed,\n * Workers, who execute tasks, and\n * Tools, that wants to inspect the state of a task.",
       "entries": [
         {
           "args": [
@@ -3091,15 +3059,15 @@ module.exports = {
         }
       ],
       "serviceName": "queue",
-      "title": "Queue API Documentation",
-      "version": 0
+      "title": "Queue API Documentation"
     },
     "referenceUrl": "https://references.taskcluster.net/queue/v1/api.json"
   },
   "QueueEvents": {
     "reference": {
-      "$schema": "http://schemas.taskcluster.net/base/v1/exchanges-reference.json#",
-      "description": "The queue, typically available at `queue.taskcluster.net`, is responsible\nfor accepting tasks and track their state as they are executed by\nworkers. In order ensure they are eventually resolved.\n\nThis document describes AMQP exchanges offered by the queue, which allows\nthird-party listeners to monitor tasks as they progress to resolution.\nThese exchanges targets the following audience:\n * Schedulers, who takes action after tasks are completed,\n * Workers, who wants to listen for new or canceled tasks (optional),\n * Tools, that wants to update their view as task progress.\n\nYou'll notice that all the exchanges in the document shares the same\nrouting key pattern. This makes it very easy to bind to all messages\nabout a certain kind tasks.\n\n**Task specific routes**, a task can define a task specific route using\nthe `task.routes` property. See task creation documentation for details\non permissions required to provide task specific routes. If a task has\nthe entry `'notify.by-email'` in as task specific route defined in\n`task.routes` all messages about this task will be CC'ed with the\nrouting-key `'route.notify.by-email'`.\n\nThese routes will always be prefixed `route.`, so that cannot interfere\nwith the _primary_ routing key as documented here. Notice that the\n_primary_ routing key is always prefixed `primary.`. This is ensured\nin the routing key reference, so API clients will do this automatically.\n\nPlease, note that the way RabbitMQ works, the message will only arrive\nin your queue once, even though you may have bound to the exchange with\nmultiple routing key patterns that matches more of the CC'ed routing\nrouting keys.\n\n**Delivery guarantees**, most operations on the queue are idempotent,\nwhich means that if repeated with the same arguments then the requests\nwill ensure completion of the operation and return the same response.\nThis is useful if the server crashes or the TCP connection breaks, but\nwhen re-executing an idempotent operation, the queue will also resend\nany related AMQP messages. Hence, messages may be repeated.\n\nThis shouldn't be much of a problem, as the best you can achieve using\nconfirm messages with AMQP is at-least-once delivery semantics. Hence,\nthis only prevents you from obtaining at-most-once delivery semantics.\n\n**Remark**, some message generated by timeouts maybe dropped if the\nserver crashes at wrong time. Ideally, we'll address this in the\nfuture. For now we suggest you ignore this corner case, and notify us\nif this corner case is of concern to you.",
+      "$schema": "/schemas/common/exchanges-reference-v0.json#",
+      "apiVersion": "v1",
+      "description": "The queue service is responsible for accepting tasks and track their state\nas they are executed by workers. In order ensure they are eventually\nresolved.\n\nThis document describes AMQP exchanges offered by the queue, which allows\nthird-party listeners to monitor tasks as they progress to resolution.\nThese exchanges targets the following audience:\n * Schedulers, who takes action after tasks are completed,\n * Workers, who wants to listen for new or canceled tasks (optional),\n * Tools, that wants to update their view as task progress.\n\nYou'll notice that all the exchanges in the document shares the same\nrouting key pattern. This makes it very easy to bind to all messages\nabout a certain kind tasks.\n\n**Task specific routes**, a task can define a task specific route using\nthe `task.routes` property. See task creation documentation for details\non permissions required to provide task specific routes. If a task has\nthe entry `'notify.by-email'` in as task specific route defined in\n`task.routes` all messages about this task will be CC'ed with the\nrouting-key `'route.notify.by-email'`.\n\nThese routes will always be prefixed `route.`, so that cannot interfere\nwith the _primary_ routing key as documented here. Notice that the\n_primary_ routing key is always prefixed `primary.`. This is ensured\nin the routing key reference, so API clients will do this automatically.\n\nPlease, note that the way RabbitMQ works, the message will only arrive\nin your queue once, even though you may have bound to the exchange with\nmultiple routing key patterns that matches more of the CC'ed routing\nrouting keys.\n\n**Delivery guarantees**, most operations on the queue are idempotent,\nwhich means that if repeated with the same arguments then the requests\nwill ensure completion of the operation and return the same response.\nThis is useful if the server crashes or the TCP connection breaks, but\nwhen re-executing an idempotent operation, the queue will also resend\nany related AMQP messages. Hence, messages may be repeated.\n\nThis shouldn't be much of a problem, as the best you can achieve using\nconfirm messages with AMQP is at-least-once delivery semantics. Hence,\nthis only prevents you from obtaining at-most-once delivery semantics.\n\n**Remark**, some message generated by timeouts maybe dropped if the\nserver crashes at wrong time. Ideally, we'll address this in the\nfuture. For now we suggest you ignore this corner case, and notify us\nif this corner case is of concern to you.",
       "entries": [
         {
           "description": "When a task is created or just defined a message is posted to this\nexchange.\n\nThis message exchange is mainly useful when tasks are scheduled by a\nscheduler that uses `defineTask` as this does not make the task\n`pending`. Thus, no `taskPending` message is published.\nPlease, note that messages are also published on this exchange if defined\nusing `createTask`.",
@@ -3636,15 +3604,14 @@ module.exports = {
       ],
       "exchangePrefix": "exchange/taskcluster-queue/v1/",
       "serviceName": "queue",
-      "title": "Queue AMQP Exchanges",
-      "version": 0
+      "title": "Queue AMQP Exchanges"
     },
     "referenceUrl": "https://references.taskcluster.net/queue/v1/exchanges.json"
   },
   "Secrets": {
     "reference": {
-      "$schema": "http://schemas.taskcluster.net/base/v1/api-reference.json#",
-      "baseUrl": "https://secrets.taskcluster.net/v1/",
+      "$schema": "/schemas/common/api-reference-v0.json#",
+      "apiVersion": "v1",
       "description": "The secrets service provides a simple key/value store for small bits of secret\ndata.  Access is limited by scopes, so values can be considered secret from\nthose who do not have the relevant scopes.\n\nSecrets also have an expiration date, and once a secret has expired it can no\nlonger be read.  This is useful for short-term secrets such as a temporary\nservice credential or a one-time signing key.",
       "entries": [
         {
@@ -3725,14 +3692,14 @@ module.exports = {
         }
       ],
       "serviceName": "secrets",
-      "title": "TaskCluster Secrets API Documentation",
-      "version": 0
+      "title": "TaskCluster Secrets API Documentation"
     },
     "referenceUrl": "https://references.taskcluster.net/secrets/v1/api.json"
   },
   "TreeherderEvents": {
     "reference": {
-      "$schema": "http://schemas.taskcluster.net/base/v1/exchanges-reference.json#",
+      "$schema": "/schemas/common/exchanges-reference-v0.json#",
+      "apiVersion": "v1",
       "description": "The taskcluster-treeherder service is responsible for processing\ntask events published by TaskCluster Queue and producing job messages\nthat are consumable by Treeherder.\n\nThis exchange provides that job messages to be consumed by any queue that\nattached to the exchange.  This could be a production Treeheder instance,\na local development environment, or a custom dashboard.",
       "entries": [
         {
```

djmitche:
That diff looks fine.  The undefined's are baseUrls, and we don't have baseUrls anymore.  Pete is rewriting client generation, so it's not worth fixing the comments.

